PR #5, opened 2018-07-30
Shannon/ad fade

smarie33:
- create fade in / fade out for rail ads

- navigate to any post page, like: http://localhost:8080/janes-embeds-test-post-1625982
- watch the ad in the right rail for 30 seconds to watch it fade out for 1 second and fade in for 1 second
- to speed this up, modify the rail component to rotate ads less that 30 seconds
  - navigate to components > Rail > index.js > line 27

- i feel like 2 seconds is too long, so i made fades 1 second each, (which together is 2 seconds). let me know if you want to 2 seconds

https://app.asana.com/0/543787180217047/759556625512872

* [ ] add a reviewer
* [ ] run appropriate tests


components/Ad/index.scss
```diff
@@ -2,3 +2,14 @@
   display: flex;
   justify-content: center;
 }
+
+.Rail {
```
andrewb: We don't want to limit fading to rail ads. That's our only use now, but there's nothing stopping us from setting a refresh interval for any ad, in which case we'll want it to fade as well.

components/Ad/index.js
```diff
@@ -1,30 +1,46 @@
...
 import './index.scss';
 
 class Ad extends Component {
   constructor(props) {
     super(props);
     this.setRefreshTimeout = this.setRefreshTimeout.bind(this);
+    this.slotDelayFade = this.slotDelayFade.bind(this);
     this.slotRefreshTimeout = '';
+    this.slotDelayFadeTimer = '';
+    this.state = { showAd: false };
```
andrewb: I think we should reverse this logic so it's `fadeAd` instead. At the moment ads that don't have a `refreshInterval` will have a `showAd` value of `false` which could get confusing.

components/Ad/index.scss
```diff
@@ -2,3 +2,14 @@
...
+  .Ad {
+    opacity: 0;
+    transition: 1s;
+    &--show {
+      opacity: 1;
+      transition: 1s;
```
andrewb: I think we should play it safe and only transition on `opacity`.

components/Ad/index.js
andrewb: That way `.Ad` can have a default opacity of `1` as well.

components/Ad/index.js
```diff
@@ -1,18 +1,27 @@
...
 import { Bling as GPT } from 'react-gpt';
+import classnames from 'classnames';
 import { getSizeProp } from './helpers';
 import './index.scss';
 
 class Ad extends Component {
   constructor(props) {
     super(props);
+
+    if (this.props.refreshInterval) {
```
andrewb: Nitpick, you could simplify to `this.state = { hideAd: !!this.props.refreshInterval }`

components/Ad/index.js
```diff
@@ -1,18 +1,27 @@
...
+      this.state = { hideAd: true };
+    } else {
+      this.state = { hideAd: false };
+    }
     this.setRefreshTimeout = this.setRefreshTimeout.bind(this);
+    this.slotDelayFade = this.slotDelayFade.bind(this);
```
andrewb: Nitpick, would a better name for this be `this.setFadeTimeout`?

components/Ad/index.js
```diff
@@ -1,18 +1,27 @@
...
     this.handleOnBlur = this.handleOnBlur.bind(this);
     this.handleOnFocus = this.handleOnFocus.bind(this);
     this.startTimers = this.startTimers.bind(this);
     this.clearTimers = this.clearTimers.bind(this);
     this.slotRefreshTimeout = undefined;
+    this.slotDelayFadeTimer = undefined;
```
andrewb: Nit, I'd call this `slotFadeTimeout` so it's consistent with `slotRefreshTimeout`.

components/Ad/index.js
smarie33: 2079942

PR #6, opened 2018-07-30
Henry/bundle voltron

chenryhen:
PR folds StackBundle into the SpreadBundle component. It also makes the picture options and css on oxygen-lib-ui much more generic. There is a companion `oxygen` PR that uses this new single component.

As you can see I broke the SpreadBundle tests. It looks like the `teasers` variable isn't being generated properly, and I am guessing it's because I didn't mock something, but I am not sure what that might be.

I went back to the more general `createSetBundleItemProps` of the StackBundle instead of the more focused `getTeaserData` of the SpreadBundle, because I was unsure of how to handle that.


chenryhen:
Oxygen PR here https://github.com/apartmenttherapy/oxygen/pull/376

PR #7, opened 2018-07-31
Henry/recipe search bundle

chenryhen:
Creates a new `Search` component that displays a `SearchForm` and quick links that are passed in as props. The `onSearch` function for the form is passed through `Search`, so we can do things like use Router in the Kitchn app.

Has a accompanying Oxygen PR


chenryhen:
@smarie33 done!

PR #10, opened 2018-08-01
Shannon/ad no timer out of viewport

smarie33:
- checks if window is not in focus and stops timers if so
- checks if window is in focus and restarts timers if so

- navigate to any post page (http://localhost:8080/whole30-approved-foods-whole-foods-260963?gentest=true)
- Right click on the ad in the right rail, wait 30 seconds and watch the `data-google-query-id` update
- click on a new tab or open a new window and see the ad no longer update

For faster testing:
- go to components > Ad > index.js line 54
- replace `this.props.refreshInterval` with 5
- see ad slot update every 5 second

- see that all tests for the Ad component pass

#### Deployment Notes

https://app.asana.com/0/543787180217047/759556625512875

* [ ] add a reviewer
* [ ] run appropriate tests


components/Ad/index.js
```diff
@@ -8,23 +8,57 @@ class Ad extends Component {
   constructor(props) {
     super(props);
     this.setRefreshTimeout = this.setRefreshTimeout.bind(this);
+    this.handleOnBlur = this.handleOnBlur.bind(this);
+    this.handleOnFocus = this.handleOnFocus.bind(this);
+    this.startTimers = this.startTimers.bind(this);
```
andrewb: Nitpick, but I'd call these `startTimer` and `clearTimer` since we only start/stopping one timeout.

components/Ad/index.js
```diff
@@ -8,23 +8,57 @@ class Ad extends Component {
...
+    this.clearTimers = this.clearTimers.bind(this);
     this.slotRefreshTimeout = '';
```
andrewb: Nitpick, I'd use `undefined` here instead of the falsy `''`

components/Ad/index.js
```diff
@@ -8,23 +8,57 @@ class Ad extends Component {
...
+    this.currentSlot = '';
```
andrewb: `undefined` here too.

components/Ad/index.spec.js
```diff
@@ -72,4 +72,26 @@ describe('<Ad />', () => {
...
+    jest.runOnlyPendingTimers();
+    const { wrapper } = setup();
+    const spyBlur = jest.spyOn(wrapper.instance(), 'handleOnBlur');
+    wrapper.instance().clearTimers = jest.fn();
+    window.addEventListener('blur', spyBlur);
+    window.dispatchEvent(new FocusEvent('blur'));
+    expect(wrapper.instance().clearTimers).toHaveBeenCalledTimes(1);
+  });
+
+  it('should restart timer if user not on window', () => {
```
andrewb: `should restart timer if user focuses window`?

components/Ad/index.js
```diff
@@ -8,23 +8,57 @@ class Ad extends Component {
...
+
+  startTimers(slot) {
+    this.slotRefreshTimeout = setTimeout(() => {
+      GPT.refresh([slot]);
+    }, this.props.refreshInterval * 1000);
+  }
+
+  clearTimers() {
+    clearTimeout(this.slotRefreshTimeout);
+    this.slotRefreshTimeout = '';
```
andrewb: `undefined`

components/Ad/index.js
andrewb: Oh, looks like there might be another timeout for fade? Going to look at that PR now...

components/Ad/index.spec.js
smarie33: 1231faa

PR #13, opened 2018-08-02
add classname for placement

chenryhen:
Courtesy PR/ Add a classname to the post bundle wrapper (instead of the item <div> itself). This feels more appropriate as far as what it being styled.


chenryhen:
looking at the tests now

PR #17, opened 2018-08-06
Add LibUi config

andrewb:
This PR adds a "top level" `LibUi` class that exposes a `configure` and `getItemComponentMap` static methods. The motivation is to simplify how bundle items overrides are handled. This change makes it possible for the consumer to configure lib ui once, and have that config used everywhere it is required. See the companion [oxygen PR](https://github.com/apartmenttherapy/oxygen/pull/404) to see how this works.

Notable changes:
- Bundle components no longer have to map or merge component overrides
- Bundles can render ANY type of component since the component map is shared. This is how I imagined things working originally. Bundles should be relatively dumb and it's up to the tooling (e.g. workflow) to enforce constraints and our styles to handle known cases.

I think we'll keep iterating here, but I feel like this is a good step.


blech75:
ooooh... this looks cool, thanks! lemme dive in to wrap my head around it...

components/index.js
```diff
@@ -0,0 +1,39 @@
...
+    image: Figure,
+    profile: PersonProfile,
+    recipe: Recipe,
+    teaser: Teaser,
+    thirdParty: ThirdPartyEmbed,
+    video: Video
+  }
+};
+
+class LibUi extends Component {
```
blech75: minor organizational nit here: the class is called `LibUi` and `configure` can ostensibly handle more than just components. so why put it in the `components` dir? shouldn't it be one level up?

components/ItemBundle/index.js
```diff
@@ -1,32 +1,18 @@
...
-import Ad from '../Ad';
 import './index.scss';
 
-// key is bundle item type, value is component
-export const COMPONENTS = {
-  ad: Ad
-};
-
 /**
  * handles rendering an ad. pretty simple/boring
```
blech75: oof.... looks like this comment could use some, err... updating or deleting.

components/index.js
```diff
@@ -0,0 +1,39 @@
...
+  static configure(overrides) {
```
blech75: it might be nice to add some jsdoc here for the class and its static methods. not a blocker, though.

components/ItemBundle/index.js
blech75: dumb question (that could probably be answered by a good comment): what _is_ the purpose of the `ItemBundle` anyway? :-)

components/index.js
andrewb: 👍 will update

components/ItemBundle/index.js
andrewb: @blech75 will do.

> A basic bundle - it simply renders a collection of items inside a wrapper. A typical use case is rendering an ad, video, or interrupter bundle that does not require the complexity of a spread.

PR #43, opened 2018-08-17
allow tags to make it through to teaser component

blech75:
courtesy PR to add 'tags' field to `BASE_TEASER_PROPS` so it makes it through to the component


chenryhen:
yeah booooooy

PR #44, opened 2018-08-20
Henry/amp

chenryhen:
adds an AMP-specific site header component


chenryhen:
ah forgot tests. coming right up...

andrewb:
@blech75 could you take over review for this if you're free?

components/AmpSiteHeader/index.spec.js
```diff
@@ -0,0 +1,38 @@
+import createSetupFunction from '../../jest/createSetupFunction';
```
blech75: you _should_ be able to just use `'jest/createSetupFunction'` since the babel aliases are set.

PR #54, opened 2018-08-23
Henry/amp in figure

chenryhen:
So this has a companion PR (https://github.com/apartmenttherapy/oxygen/pull/499) that I'm actually more concerned about. This adjusts some component props for the new state of affairs re: images, and uses a set container size for AMP images in `Figure`


chenryhen:
I'll fix those tests

chenryhen:
Closing due to new (non-)AMP strategy

PR #69, opened 2018-08-28
igor/spot-im-fix-tests: Fix warnings in tests related to spot-im

demo-igor:
Fix warnings in tests related to spot-im


components/SpotIM/index.spec.js
```diff
@@ -1,7 +1,12 @@
 import createSetupFunction from '../../jest/createSetupFunction';
 import SpotIM from './index';
 
-const setup = createSetupFunction(SpotIM, {});
+const MOCK_DATA = {
```
blech75: this is fine and works a-ok. just a minor style nit...

i've typically just inlined the props object unless i need to refer to it later on:
```
const setup = createSetupFunction(SpotIM, {
  spotId: 'SPOT_ID',
  postId: '12345678',
  canonical: 'asd'
});
```
(that and it's not "mock data" per se, it's just the "default props" for the test instances.)

demo-igor: Updated code. @blech75  could you please review one more time?

blech75:
@demo-igor thanks... good to go! :shipit:  (it was already approved without the changes.)

PR #81, opened 2018-09-04
Henry/image credits

chenryhen:
Courtesy PR. WP only creates a caption, not a separate credits field, so the components have been simplified to only show one piece of content along with an image. Not touching the schema right now as everything (including Flow, I believe) is still set up for an additional credit field.


chenryhen:
forgot to add: https://app.asana.com/0/699007310865935/802668187679627/f

PR #88, opened 2018-09-07
Shannon/cloudinary dpr

smarie33:
- change images using cloudinary to use dpr
   - dpr automatically selects the proper image size per browser dpi
- remove dpi calculations from resolution list for cloudinary fetched images

- also use pr: https://github.com/apartmenttherapy/oxygen/pull/575
- desktop browser
  - navigate to http://localhost:8080/collection/bundle-samples
  - test in safari and chrome at different browser sizes (refresh per resize because dpr image is delivered on page load)
  - inspect any image
  - hover over the `img > src` and see the dimensions of image delivered is the close to the image displayed per media query for chrome and around 2.5x the size for safari on retina screens and around 1x on non retina screens.
- iphone
  - connect iphone to computer via phone usb
  - open up network utility
  - under info tab see `IP Address`
  - type this address into your iphone's safari + `:[localhost port number]/` + the rest of the url above
  - open safari on your desktop
  - under the Develop tab select iphone > safari the url that you went to on your phone (should be the ip address)
  - drill down the dom to an image and see that the image dimensions delivered is around 2.5x the size of the image displayed

#### Deployment Notes
also has a oxygen pr: https://github.com/apartmenttherapy/oxygen/pull/575

https://app.asana.com/0/543787180217047/801498898904057

* [ ] add a reviewer
* [ ] run appropriate tests


chenryhen:
I'm getting the correctly sized images, but I'm wonder if dpr works in the context of <picture>? It looks like the browser is still making a decision based on the the various <source> `srcset` values. Or maybe it's because the urls in the <source> don't have the dpr setting in the urls.

If we _do_ continue to use dpr, @andrewb what do you think of just adding it to the `DEFAULT_OPS` in media.js? We seem pretty sure of Cloudinary for the time being.

andrewb:
@smarie33 @chenryhen I don't think `dpr_auto` is an option for us yet. It looks like it relies on client hints which is not widely supported (https://caniuse.com/#search=client%20hints). Unless I'm mistaken I think our best bet is to keep the source sizes we have (i.e. no exception for Cloudinary), and just tweak the sizes (like in the other PR).

smarie33:
@andrewb @chenryhen Ok. So I'm closing this pr bc it just adds dpr stuff. the corresponding lib-ui pr changes images sizes throughout the teasers to make sure they are the correct sizes per media query

PR #92, opened 2018-09-12
revise image placeholder approach for posts

blech75:
this PR completely revises the approach to how we render "placeholder" images in posts (normal image embeds, not galleries).

previously, we used a placeholder element _inside_ the lazy-load container. this did not perform well during rendering/scrolling for a confluence of reasons: nodes were being swapped and image element dimensions were unknown. setting a placeholder element to an assumed height was not good enough; the node was removed before you saw it, and the latency of requesting the hires image _of unknown height_ (until the image header arrived) caused the browser to aggressively reflow the layout multiple times. no good. :-(

we now use a specialized placeholder image that is ~10% or less of the original image size. it's present in the original markup so that the browser can immediately start downloading it. this allows the browser to then set layout on the elements to the final size. thanks to some clever CSS, we then "overlay" the final image directly on top of the placeholder.

this placeholder image uses `f_png,w_768,e_pixelate:100` params, yielding an image that's about 1/10 the size (or less) than the original. ex:

(i chose the pixellation effect so that the filesize was more consistent. we can easily alter the params to use a different "style" image.)

note that you don't see the hires render progressively as the image loads in. my sense is that because the hi-res is `position: absolute`, the layout engine does not display the image until its full dimensions are known.

oh, and this also changes the "pin it" button so that it's a child of the lazy load component -- it will only appear over the final hi-res image (not the placeholder, which was kinda weird).

@chenryhen please look at this and let me know what you think. i have some hesitation about the `placeholderHeight` parameter that i calculate because i'm not entirely sure about what the real data looks like. also note that hires images are now `position: absolute`, something i don't particularly love. i had to tweak the "pin it" button for this. can you think of other bits that are sensitive to `position: absolute`?

finally: this is mainly a stopgap until we can get actual image dimensions data in the bundle. not having the dimensions means we _need_ the browser to get _some rendering_ of the image in order to determine the height/layout.

this is a start, but there remains a few ways to further improve this:

* **set `onload` handler for placeholder image** that updates the props of the lazyload component to set a `height` prop (and lazyload is issuing warnings). my guess is that this _should_ alleviate the scrolling-up issue i mention in the comments.
* **remove the original placeholder image from the DOM**. i don't live the fact that our placeholder image remains and that our final image is `position: absolute`. this will require a "smart" container that updates the class on the wrapping element to indicate the "placeholder mode/phase" is complete. we'd update the CSS so that the `position: absolute` stuff for the hires image is contextual to being in "placeholder mode". we'd probably also want to add `height` and `width` props to the hires `img` element for good measure, too.
* **add a clearer indication that it's a placeholder image**. i already set the `opacity` on it to hint that it's a placeholder. we could animate the opacity to create a subtle "throbbing" effect, or we could add a "loading" indicator via `::after` or something.
* **reduce file size of placeholder image even more**. there's a few limiting factors here: image format, some CSS layout assumptions for smaller dimension images, etc.

of course, i'd like to run some tests to get absolute metrics on how this affects page load (time, bytes, CPU, etc.), but i think might have to hold off that for now....


blech75:
see this report for placeholder vs. hires image:

screen shot 2018-09-12 at 11 04 53 am

components/Figure/index.js
```diff
@@ -40,35 +41,74 @@ function Figure(props) {
     ops: {
       // Crop image if width and height are set
       ...(width && height && { crop: 'crop' })
-    }
+    },
+    className: 'Figure__image-hires'
```
chenryhen: BEM-wise, this should be `Figure__image--hires`

components/Figure/index.js
```diff
@@ -40,35 +41,74 @@ function Figure(props) {
...
+  // the max width of the image in the layout is currently 730px. our maxWidth
+  // is currently 1536px. in order to minimize the file size of the placeholder
+  // image, we halve its dimensions because we don't need it to be hi-dpi. this
+  // assumes, of course, that we request all of our images in their hi-dpi size.
+  //
+  // CLEANUP: verify above assumption
+  // CLEANUP: might want to extract this to a separate method/module
+  //
+  const placeholderWidth = Math.round((maxWidth || originalWidth) / 2);
+  const placeholderHeight = originalHeight
```
chenryhen: Do we actually need the height? If we're not changing the aspect ratio, i.e. cropping, of the image, can Cloudinary return a scaled up/down version just based on width?

components/Figure/index.js
blech75: gah, sorry! will change.

components/Figure/index.js
blech75: ok, so we're not changing the aspect ratio _here_ (for posts). however, i was planning on extracting this for use by teasers. and if we do get a height passed, i think we need this. does that make sense? (my head still hurts thinking about all this stuff....)

components/Figure/index.js
blech75: updated both classes. see 9235e5c

components/Figure/index.js
chenryhen: 10-4

chenryhen:
GTG :shipit:

blech75:
note: we talked about a possible refactor in the future to move from the `position: absolute` approach to using CSS grid instead.

PR #103, opened 2018-09-17
Henry/comment count

chenryhen:
We want to add a comment icon/count to the post actions. Because Spot.IM is lazy loaded, we can't use the same implementation as ATCMS (where the widget hooks into a class name). Instead we are using the open API. So, now `PostBundle` is class component, and we query for the comment count when it's loaded on the client side.

@blech75 I would very much appreciate some help with the tests for the new functionality.


chenryhen:
whomp whoooooomp. At least I know how to do it now.

blech75:
yeah, what @andrewb said... any data fetching should not be in lib-ui.

also, i'd personally try to factor out the comment fetching to a separate "helper" method to facilitate testing (something named `getCommentCountForPost()`). that way, your spec for `PostBundle` is purely looking at whether a call to `getCommentCountForPost()` is made on mount. the helper can be more specific about testing for `fetch()`.

FWIW, i took one way of testing data fetching in [`EmaiListDataProvider`](https://github.com/apartmenttherapy/oxygen/blob/master/kitchn/components/EmailListDataProvider/index.spec.js#L99-L130) and @andrewb took another way in [`Recirc`](https://github.com/apartmenttherapy/oxygen/blob/master/kitchn/components/Recirc/index.spec.js#L68-L148).

Or, instead of a helper method, you could taking the same path I took in `EmaiListDataProvider`, and make a `CommentCountProvider` component that wraps the UI for displaying the UI bit you need.

components/PostActions/index.js
```diff
@@ -27,16 +36,38 @@ function PostActions(props) {
...
+            <div className="Post__comment-count">{commentCount}</div>
+          )}
+        </a>
+      </div>
     </div>
   );
 }
 
 PostActions.defaultProps = {
+  commentCount: undefined,
```
blech75: hmmm... i dunno, i _might_ suggest making the default `0` only so it reads better. i know `undefined > 0` evaluates to `false`, but, ehhh.... just rambling.

chenryhen: No, that's fine. This is the sort of thing where I don't know best practices or what kind of style of coding we want.

components/PostActions/index.js
```diff
@@ -1,10 +1,19 @@
 import React from 'react';
-import { string } from 'prop-types';
+import { number, string } from 'prop-types';
 import ShareAction from './ShareAction';
+import { Comment } from '../../constants/icons';
```
blech75: i typically mock the `Comment` component in the spec. speaking of which, you could add specs testing for `commentCount` if you feel like it, but it's not totally necessary since the logic is 'simple'.

PR #120, opened 2018-09-25
add image services + images helper

blech75:
introducing....

* `helpers/images`: the new place to get your image URLs. (`mediaUrl` is still being used until  a future PR deprecates/removes it.)
* `lib/imageServices`: the new abstraction over per-service image URLs. (just in time for deprecating imgix!!!)


blech75:
FWIW, i flubed a few things here:

* missed setting a default quality
* screwed up the crop/fill param

fixing that now.

PR #129, opened 2018-09-27
Shannon/quantcast

smarie33:
- change jwplayer events to run multiple functions on trigger

- please see pr: https://github.com/apartmenttherapy/oxygen/pull/669 for testing
- also run lib-ui tests

#### Deployment Notes

https://app.asana.com/0/543787180217047/791966078829326

* [ ] add a reviewer
* [ ] run appropriate tests


components/JWVideo/helpers.js
```diff
@@ -54,33 +54,69 @@ export const JW_EVENT_NAMES = [
  * @param {Function} customHandler a custom event handler for the event
```
blech75: psst. jsdoc needs to be updated.  you're passing an array of functions now. :-)

blech75:
i was going to start leaving some comments on the code, but i wanted to take a step back and make a general comment on the possible approaches to the problem as i see it. i could be slightly off-base, but bear with me as i think through this...

handling multiple methods per event could be done at least two ways:

* modify `createEventHandlers()` to handle accepting _multiple sets of events_ that could (and do) have the same event names. this is what you have now; it requires re-working `JWVideo/helpers.js` a fair amount.

* pre-compose/combine the multiple sets of event handlers into a single set. this allows you to leave the existing bits of `JWVideo/helpers.js` alone. the idea is to keep the event combining/wrapping logic separate from the already-functioning event handler bits. (whether or not the combining logic would go into `JWVideo/helpers.js` or the site-specific component could be up for debate.)

what do you think of the pros/cons? can you envision a way to pre-compose/combine the event handlers before passing them into `JWVideo`?

blech75:
@smarie33 i don't like or feel good about doing this, but i'm going to close/deny this PR in favor of another one uses the `aggregateEvents()` function i showed you. see #189.

the primary reasons for this is due to the unnecessary alteration of existing code and the addition of new code without tests. this event stuff is sufficiently complicated and i'd prefer to keep known/tested code working.

#189 will require a minor change to apartmenttherapy/oxygen#669:
```js
const customEventHandlers = JWVideo.aggregateEvents(quantcast, comscore);
```

if you have any questions about this, please let me know!

PR #134, opened 2018-09-28
simplify responsive images codepath

blech75:
courtesy PR. doing a bunch of refactoring here, that should result in NO LOGIC CHANGES. i diff'd rendered `<picture />` and `` tags when using this branch vs. not, and they both appear identical.

the idea here was to remove the `buildSources()` wrapper because all that did was fork based on if the image is 'fixed size' or not. 

the notion of am image being 'fixed size' has changed from a data prop to a dynamic check on the units. with that in place, we could easily rework and simplify the code path.

with that in place, it demonstrates that an `` always receives the result of `generateWidthsList()`. that is _not_ obvious in the previous incarnation.


blech75:
i merged this too quickly. there's a minor issue with gallery images. getting a minor warning in Gallery:

`Warning: Failed prop type: The prop 'sizes[0].dimensions.width' is marked as required in 'Placeholder', but its value is 'undefined' in Placeholder`

fixing this now....

blech75:
ok, nevermind. i verified things on the live site and it's exactly the same. just a dev mode vs. prod mode thing. that said, we should resolve this.

the good news is we're gonna be revising how placeholders work, so it'll be fixed shortly.

PR #141, opened 2018-10-02
refactor spread and picture

blech75:

this PR revises how we describe and build `<picture>` and `<source>` elements with some minor prop tweaks to `SpreadBundle` in the process.

* a `SpreadBundle` now accepts `bundlePictureSpecs` (instead of `pictureOptions`), which has two keys: `base` and `exceptionPatterns` (which defines _patterns_ that are later are expanded to _actual exception positions_).
* `base` describes the default/base way `<picture>` are rendered in the bundle. it contains
  * `fallbackDims` (`{ width, height }`)
  * `sourceSizes` (see below)
* `sourceSizes` is the data structure that is used to generate the value for `srcSizes`, so it has three keys:
  * `media` (media query)
  * `imageDims` (`{ width, height }`)
  * `layoutWidth` (`{ value, units }`)
* `exceptionPatterns` is the same idea as before, but updated for the above. it contains:
  * `positionStart` and `recurrence`, both unchanged
  * `fallbackDims` and `sourceSizes`, used as described above
* `exceptionPatterns` are expanded into `execeptionPictureSpecs` (`{ idx: pictureSpecs }`), which  describes the picture specs (`{ fallbackDims, sourceSizes }`) for each exception position.
* we also rework/rename `ops` to be more specific by naming it `imgRenderOps`, and that is passed into the method that generates the final image URL and contains parameters to adjust _rendering_ of the image.
* we separate out the image dimensions (`imageDims`) from the `renderOps` to make logic easier to understand. this will make generating placeholders easier.

highlights:

* 6581567
  * introduce `bundlePictureSpecs` prop (replacement for `pictureOptions`)
  * rename `assignExceptionPictureOps()` to `extrapolateExceptions()`
  * rename `exceptions` to `.exceptionPatterns`
  * rename `exceptionPositions` to `exceptionPictureSpecs`
  * mock `Teaser` in specs
* 7894c92
  * extract `<source>` logic to new `PictureSource` component
  * update components to use new props: `sourceSizes`, `fallbackDims`, `imageDims`, `imgRenderOps`
  * factor out `sourceSizeProps` to `responsiveImages` module
  * rename/refactor `calculateMaxSize()` → `calculateMaxWidth()`
  * rename `expectedWidth` to` layoutWidth`
  * update `Teaser` for new prop names: `pictureOps` → `sourceSizes` + `fallbackDims`
* 70af0ea
  * remaining `ops` vars are now named `renderOps` to indicate that they

NOTE: this needs to be tested in conjunction with the related oxygen PR. (link TK)

* view `SpreadBundles` in their natural habitat: home page, collection pages, recirc bundles, etc
* ensure that the markup is identical to the existing markup

* does the naming make sense? is it an improvement? (🤞)

#### Deployment Notes

n/a

* [ ] add a reviewer
* [ ] run appropriate tests


chenryhen:
Did a spot check and the picture sources look like they match what's on production right now :+1:

components/Picture/source.js
```diff
@@ -0,0 +1,72 @@
...
+};
+
+/**
+ * handles rendering an individual <source> element for use within a <picture>.
+ *
+ * decides which approach to use for the srcSet based on if the provided image's
+ * layoutWidth is specified in fixed vs. relative units.
+ *
+ */
+function PictureSource(props) {
```
andrewb: Nit, I think this should be `/PictureSource/index.js`

blech75: yeah, i hear you. i'm torn between separate component or sub-component in situations like this. lemme think about it again.

blech75:
so after all this time, i still managed to screw up the naming. `sourceSizes` should be `sourceSets`. i'm going to fix this... soon.

PR #151, opened 2018-10-04
add more placeholders

blech75:
placeholders have arrived! 🎉

* figures are slightly optimized
* galleries now have low-res placeholders similar to figures (pixelated)
* all teasers utilize the a light blue pattern:
https://cdn.apartmenttherapy.info/image/fetch/f_auto,q_auto:eco,c_scale,w_200,h_250/l_placeholders:tile-light-blue,fl_tiled/https://s3.amazonaws.com/pixtruder/original_images/dd436318cb8c08bf7d17d5838e06e296e0a25e35

review:

* galleries + figures: http://localhost:8080/shrimp-boil-recipe-259082
* teasers: homepage + any/all collection pages + recircs at bottom of posts

review with bandwidth throttling -- i've been using 5mbps, 10mbps, 20mbps --and  filter on network requests containing `cdn.apart`. look for PNG images that are tiny (way < ~10kb) and original images that are webp or jpeg (and big... many 10s of kb or 100s of kb). the placeholder PNG requests should come BEFORE the hires original formats.

**UPDATE:** PNG images are still used for the post placeholders because they are still pixelated. the teaser placeholders are now using auto (WebP/JPEG) because the pattern compresses better (smaller file size) than in PNG. filter network requests for `dd43`.

there should be substantially less reflow. ~~that said, we still have some issues to work out. the most noticeable are:~~ → **ALL ISSUES FIXED!**

* ~~small viewport sizes. @chenryhen this needs some CSS review everywhere. especially at smaller viewport widths, where i know the grid layout is slightly busted for things like the big soundbar.~~ → FIXED in db6ef4a

* ~~the teaser block relies on an image to "prop open" the grid area. we know the aspect ratio of the image that fits in the teaser, but not the physical dimensions. we need something in the teaser that will render to the correct size to "prop open" the teaser. perhaps we can leverage the existing `Placeholder` element to do the job, always keeping the element in place. i dunno; need to experiment.~~ → resolved. we're now loading a single common image for all teasers of a certain size.

* ~in theory, it might be nice to force the teaser to _always load the placeholder first_ (since it's so small). but that would require a callback to initialize the loading of the original image. this is similar to but orthogonal to the concept of lazy loading components, and i don't want to make things more complicated than they need to be. this is most visible when doing client-side navigation to new pages, though it happens in any lazy-loaded bundle situation  the (new) bundles render in-viewport (or within the offset), so lazyload loads the original element/image immediately, which therefore competes for bandwidth with the placeholder image. sigh.~~ → resolved. we're now loading a single common image for all teasers of a certain size.

one other thing i changed was that the placeholders now only generate 1 resolution and 1 width descriptor. since they're placeholders, let's keep things simple, small, and as uncomplicated as possible.


components/TeaserImage/index.js
```diff
@@ -26,29 +26,31 @@ function TeaserImage(props) {
...
-    >
-      {picture}
-    </LazyLoad>
+    <Fragment>
+      <LazyLoad offset={LAZY_LOAD_OFFSET} once>
+        {originalImage}
+      </LazyLoad>
+      <Picture
+        className="Teaser__picture--placeholder"
+        url={url}
```
andrewb: I'd love to use the same placeholder image for all cropped images to reduce the number of requests. It won't look as pretty, but should be better for perf. I did a quick test by hardcoding `url` here to `https://storage.googleapis.com/gen-atmedia/3/2018/09/6d60bc3fbcd7d52901b892e71be399bb9d58cb0b.jpeg` and it seemed to work fine for teasers.

Of course we don't want that image, but we could design a generic image that pixelates nicely so that we don't have to branch placeholder effects.

components/TeaserImage/index.js
```diff
@@ -1,12 +1,12 @@
...
+import React, { Fragment } from 'react';
 import Picture from '../Picture';
 import AmpImage from '../AmpImage';
 import LazyLoad from '../LazyLoad';
-import Placeholder from '../Placeholder';
 import { imageUrl } from '../../helpers/images';
 
-export const LAZY_LOAD_OFFSET = 250; // px
+// CLEANUP: possibly dynamically adjust this (based on device/viewport).
+export const LAZY_LOAD_OFFSET = 50; // px
```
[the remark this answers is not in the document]
blech75: hmm.... i feel `LAZY_LOAD_OFFSET` it still is necessary to mitigate the visible rendering/reflow everywhere.

separately, i actually didn't mean for me to commit the change to 50.... i was playing around with the value.

components/TeaserImage/index.js
blech75: yeah, the generic image would definitely work performance-wise.

however, i don't love that semantically -- in the served markup, i mean -- all the images will be the same. though this is for teasers, so i guess it doesn't really matter all _that_ much.

in any case, my next step is to introspect the image dimensions of the `pictureSpecs` to make the wrapping element aware of what size (height) element it contains. that should drastically improve things.

components/TeaserImage/index.js
```diff
@@ -26,29 +26,31 @@ function TeaserImage(props) {
...
-      placeholder={<Placeholder sizes={placeholderSizes} />}
-      once
-    >
-      {picture}
-    </LazyLoad>
+    <Fragment>
+      <LazyLoad offset={LAZY_LOAD_OFFSET} once>
+        {originalImage}
+      </LazyLoad>
+      <Picture
```
andrewb: Should we pass an `alt` of "placeholder"?

blech75: yeah, i was thinking about this... _probably_, but not sure what the best practice is here. i'll read up on it.

two other related things:

* IIRC, right now the fallback `` is not actually a "placeholder" (in the low-res sense); it's the actual image.
* we have picture support and JS support as the two dimensions in our "support matrix" for this feature. i'll review the behavior in each case.

components/TeaserImage/index.js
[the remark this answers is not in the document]
blech75: @andrewb i added some comments in 8862320 to explain my rationale.

components/TeaserImage/index.js
blech75: ok, so this is now resolved. i like it. yay!

components/TeaserImage/index.js
blech75: added placeholder text  in 1b42b6c. still not entirely confident, but it's fine for now.

blech75:
@andrewb @chenryhen @smarie33 this is ready for review again when you get a chance. thx!

blech75:
@chenryhen there is one issue i want to put on our collective radar. see comment added in e717229.

blech75:
@chenryhen not sure; we can definitely test that.

blech75:
incorporated into apartmenttherapy/oxygen#764

PR #157, opened 2018-10-09
Henry/sponsor conditional

chenryhen:
Small, but this more about coding style than functionality, so I'd appreciate an extra set of eyes.

There are a lot of nested conditionals because none of the `sponsor` data is required, so I'm doing this to get the markup and wrappers that I would like for styling.  Is there a neater way to make these checks?


components/SpreadBundle/index.js
```diff
@@ -187,7 +190,7 @@ SpreadBundle.propTypes = {
     sponsor: shape({
       imageUrl: string,
       messaging: string,
-      name: string.isRequired
```
blech75: so i question the removal of `isRequired`. if `sponsor` is not required, and none of its attributes are required, then... ?

i'd prefer to have this data shape more explicitly described, and if it were a separate component (as mentioned above), then the props could be defined there and imported/reused here.

components/SpreadBundle/index.js
```diff
@@ -91,38 +91,41 @@ function SpreadBundle(props) {
...
+                    {sponsor.messaging}
+                  </span>
                 )}
-              </span>
-            </div>
-          )}
+                <span className="Spread__cake-name">
+                  {sponsor.imageUrl ? (
+                    <Image
+                      className={`Spread__cake-logo--${camelcase(
```
blech75: this is not directly related to this PR, but i worry about the `name` attribute and its exact contents. i can imagine `name` being the full, trademarked name, so we could end up with something like this:

`camelcase('Room & Board')` → `'room &Board'`.

my initial suggestion would be something like [slugify](https://github.com/sindresorhus/slugify), which _should_ remove any weird punctuation and just return words:

`camelcase(slugify('Room & Board'))` → `'roomAndBoard'`

...and it looks like we already have `slugify` as a dep, so yay.

components/SpreadBundle/index.js
```diff
@@ -91,38 +91,41 @@ function SpreadBundle(props) {
...
-                          units: `px`
-                        }
-                      }
-                    ]}
-                    alt={sponsor.name}
-                  />
-                ) : (
-                  sponsor.name
+          {title && <h4>{title}</h4>}
+          {sponsor &&
```
blech75: FWIW, i'd probably separate this sponsor block out to its own component. it's getting a little unwieldy (lots of lines and indenting) and could use some tests to better describe its behavior w/r/t choosing which elements to display, etc.

chenryhen: Good idea. I'll make a card for it.

PR #166, opened 2018-10-16
Henry/breakout sponsor

chenryhen:
Not entirely sure of this pattern, but let me know if it makes sense. This is sort of a sub-component, something that would wouldn't foreseeably exist outside another component, and whose styling is almost completely contextual. The first use case for this is adding a sponsor to a SpreadBundle (see `/recipes`, but I could see it used in an ItemBundle or on a post or Ad bundle.


components/BundleSponsor/index.js
```diff
@@ -0,0 +1,65 @@
...
+import Image from '../Image';
+import { toSlug } from '../../helpers/tags';
+
+const MAX_LOGO_SIZE = 128;
+
+// CLEANUP Capitalize the variant name to normalize it for BEM
+function BundleSponsor({ cssHook, imageUrl, messaging, name }) {
+  // We use "Cake" to avoid adblock class/id blacklists
+  const baseClass = `${cssHook}__cake`;
+  const classnames = {
```
blech75: minor nit: i'd rename this var since it conflicts mentally (for me, at least) with the `classnames` module we sometimes use.

components/BundleSponsor/index.js
```diff
@@ -0,0 +1,65 @@
...
+    name: `${baseClass}-name`
+  };
+
+  return (
+    <div className={baseClass}>
+      {messaging && <span className={classnames.messaging}>{messaging}</span>}
+      {(name || imageUrl) && (
+        <span className={classnames.name}>
+          {imageUrl ? (
+            <Image
```
blech75: nothing to _fix_ per se, but just a style note.

i've found myself using more intermediate variables. yeah, they're only used once, but it does two things: 1) makes the code/intent easier to understand since it ascribes meaning to the values, and 2) shortens up the "blocks" a bit, which makes the code easier to parse for humans. too much indenting hurts my eyes. (i see indenting is a code smell.)

components/BundleSponsor/index.js
chenryhen: totally makes sense

PR #174, opened 2018-10-17
normalize image widths

blech75:
@chenryhen let's discuss.

basically, everything (728, 768) goes to 730, because that's the column width. any multiples are also based off 730.

one thing i didn't touch was `medium4x5`. see https://github.com/apartmenttherapy/oxygen-lib-ui/blob/master/constants/imageSizes.js#L34

did i get this right? half of me says yes, the other half says no.

see https://app.asana.com/0/543787180217047/870933258710170/f


components/Figure/index.js
```diff
@@ -1,13 +1,15 @@
...
 import LazyLoad from '../LazyLoad';
 import PinItButton from '../PinItButton';
 import ImageCaption from '../ImageCaption';
 import { imageUrl } from '../../helpers/images';
 import './index.scss';
 
-export const MAX_POST_IMG_WIDTH = 1536; // px
+// CLEANUP: multiplying by the highest supported DPI. abstract this out. see
+// responsiveImages helper.
+export const MAX_POST_IMG_WIDTH = POST_COL_WIDTH * 2; // px
```
chenryhen: Why wouldn't we just import `HI_DPI_FACTOR` from `responsiveImages` here?

components/Gallery/index.js
```diff
@@ -12,8 +12,7 @@ import './index.scss';
 
 const AD_INSERTION_IDX = 4;
 
-// for the current formats, the image shouldn't ever be larger than 768px
-const MAX_GALLERY_IMAGE_WIDTH = 768;
+const MAX_GALLERY_IMAGE_WIDTH = POST_COL_WIDTH;
```
chenryhen: I think this is ok for now, but fyi the new house tour gallery will probably require some changes here. But that will probably part of a larger set of changes.

components/Image/index.js
```diff
@@ -19,12 +19,15 @@ export const sourceSizeProps = {
   }).isRequired
 };
 
+// tied to current CSS
+export const POST_COL_WIDTH = 730; // px
```
chenryhen: should this be pulled out of the component into a helper or a `/constants/` file?

components/Figure/index.js
blech75: you are correct, sir.

components/Image/index.js
blech75: yeah, i was thinking about that. wasn't sure where to put it. will poke around a bit more.

blech75:
@chenryhen should `medium4x5` get normalized, too?

chenryhen:
@blech75 no, those were based on teaser sizes. However, the spreads have gone through so many changes, it probably wouldn't be bad for me to do another audit on those sizes. Maybe as we're getting AT ready for launch

PR #177, opened 2018-10-19
we don't actually use these so unrequire them

chenryhen:
Courtesy PR to get rid of the console complaint. Also because it neither the author image nor url are required right now. We don't have author pages yet, and we don't show the profile pic (although we might. who knows!)


blech75:
thanks. that reminds me. i want to unify/centralize the prop types for data structures like this so they're not re-created all over the place. see what i did in pictures for [`sourceSetPropTypes`](https://github.com/apartmenttherapy/oxygen-lib-ui/blob/master/helpers/responsiveImages.js#L4-L19)

PR #209, opened 2018-11-15
Henry/external teaser

chenryhen:
Adds a TeaserIcons component to the structure of a teaser. When certain flags are set, the icons are used to indicate a feature of the teaser. Currently, only `isExternalLink` has an effect.


components/Teaser/icons.js
```diff
@@ -0,0 +1,26 @@
...
+
+const ATTRIBUTE_ICONS = {
+  isExternalLink: <ExternalLink />
+};
+
+function TeaserIcons({ teaserAttributes }) {
+  return (
+    <div className="Teaser__icons">
+      {teaserAttributes.map(attribute => (
+        <div className="Teaser__icon" key="attribute">
```
blech75: right now, if you pass in `hasGallery` and `isVideo`, it appears that the `div.Teaser__icon` gets output. i don't think you want that, right?

also, i think you mean `key={attribute}`.

components/Teaser/icons.js
```diff
@@ -0,0 +1,26 @@
...
+          {ATTRIBUTE_ICONS[attribute]}
+        </div>
+      ))}
+    </div>
+  );
+}
+
+TeaserIcons.propTypes = {
+  teaserAttributes: arrayOf(oneOf(['hasGallery', 'isExternalLink', 'isVideo']))
```
blech75: mostly a style thing, but i might do something like this:
```js
  teaserAttributes: arrayOf(oneOf(Object.keys(ATTRIBUTE_ICONS))
```
to more explicitly bound the allowed values and mappings together.

components/Teaser/icons.spec.js
```diff
@@ -0,0 +1,22 @@
...
+describe('<TeaserIcons />', () => {
+  describe('sanity check', () => {
+    it('should render', () => {
+      const { wrapper } = setup();
+      expect(wrapper.exists()).toBe(true);
+    });
+  });
+
+  describe('uses attributes properly', () => {
+    it('shows external link icon when `isExternalLink` flag is set', () => {
```
blech75: you should mock the icons. something like this should work.
```
jest.mock('../../constants/icons', () => ({
  ExternalLink: jest.fn()
}));
```

components/Teaser/index.spec.js
```diff
@@ -78,6 +78,15 @@ describe('<Teaser />', () => {
           .text()
       ).toEqual(cta);
     });
+
+    it('renders icons when supplied with `props.teaserAttributes`', () => {
+      const teaserAttributes = ['isExternalLink'];
```
blech75: you should mock `TeaserIcons` in this file.
```js
jest.mock('./icons');
```

components/Teaser/icons.js
chenryhen: no and yes.

PR #211, opened 2018-11-16
Henry/condensed byline

chenryhen:
Remove timestamp from the top of the post and change the wording at the bottom. Companion Oxy branch to come


blech75:
@chenryhen even though you removed `updatedAt` prop from the `PostByline` component, it's still being called with the `updatedAt` prop in the `PostBundle`:
```
<PostByline author={author} updatedAt={modifiedAt} />
```
do you envision bringing back the "last updated" wording in the byline? or anywhere on the page, for that matter?

i will remove the crufty `updatedAt` prop from that component, just say the word!

chenryhen:
@blech75 We might bring it back, but it's probably small enough that we can just add it back later? It's probably fine to take the prop out. Thanks for catching that.

blech75:
addressed in 567c82ba13fd1fc6b63f5b4a7d955de8021d8dd0

PR #216, opened 2018-11-19
make PostBundleItem stateful

blech75:
courtesy PR to make `PostBundleItem` stateful so we can set dynamic classes on the bundle item.

this is needed for the minimized right rail video.


blech75:
closing this out for now. i came up with a different approach.

PR #227, opened 2018-11-30
CSS Show/Hide Nutrition

seawolff:
+ Shows/Hides nutrition info with css only
+ Fixes test for showing / hiding nutrition

+ Confirm nutrition data is hidden by default and toggles correctly 

https://app.asana.com/0/0/927680595538552/f

* [x] add a reviewer
* [x] run appropriate tests
* [x] tag anyone who might want to see this


components/Recipe/RecipeActions.js
```diff
@@ -42,11 +42,10 @@ class RecipeActions extends Component {
             </button>
           </div>
         )}
-        {nutritionalGuide &&
-          isNutritionVisible && (
-            <div className="RecipeActions__nutrition">
-              <NutritionalGuide {...nutritionalGuide} />
-            </div>
+        {nutritionalGuide &&(
+          <div className={["RecipeActions__nutrition", (isNutritionVisible ? "" : "hide")].join(" ")}>
```
andrewb: We can use [`classnames`](https://github.com/JedWatson/classnames) here.

Here's an example in lib-ui:

https://github.com/apartmenttherapy/oxygen-lib-ui/blob/36ed3eb22522270a00fe31b010f549e27986090e/components/PostBundle/index.js#L66-L69

andrewb: We should also use BEM syntax, e.g. `RecipeActions__nutrition--hidden`.

andrewb: Alternatively, in a situation like this I would not be opposed to an inline style since it's related to logic, i.e. you could conditionally set `display: none`on the wrapper.

seawolff: Yeah I think I might go with the inline style here but thanks for the info as it does seem a bit overboard to make changes in two repos for something so small now that I think about it.

PR #235, opened 2018-12-07
cloudinary image urls

stephensauceda:
This is potentially super naive but it appears to be all that's necessary to get direct Cloudinary urls working (`cdn.apartmenttherapy.info/image/upload`). This logic might be better suited to living in `cloudinary.js` and then potentially bringing back the `testUrl` method and only having it check for the direct cloudinary url and letting the `fetch` stuff handle everything else. Kinda like having two image services but with one image service...


blech75:
heh. it's super-naive :-) , but that's ok. this image code is a little different than how we've previously approached things.

my idea was to have no service-specific logic in the `helpers/images`. that got really hairy really quickly and encouraged some wacky logic, making it very cumbersome to support different notions of render paramters - path vs. query string; different values; etc.. (if you're curious, see #144 and its linked PRs.)

instead, we'd build different service adapters (cloudinary, cloudinaryFetch, imgix, etc.), expose two methods for each service: `testUrl()` and `createUrl()`, expose all image services via `lib/imageSerices`. since we have no idea what URL is being served to us, we have to iterate over each service, and run the image URL against each `testUrl()` method to see if it matches; if so, run the `createUrl()`. that's the bit i commented out when i short-circuited it for `cloudinaryFetch` because our data was updated to now just pass us the 'storage' URLs, which we just pass to fetch.

now that we're re-introducing the notion of different kinds of URLs being passed in, we should really revert back to the looping over different services to test which one should be used.

the factoring out of fetch vs. upload was not _quite_ finished (just noticed another misleading comment) -- i had planned to revisit this when we re-introduced upload bit.

is that enough to go on? if not, we could discuss futher and/or pair on this. or i can poke at it again later. it shouldn't take _too_ long because all the plumbing is already there.

blech75:
@stephensauceda i just re-read your comment and realized that you already understand what i so tediously explained. sorry! at least it's somewhat documented now. heh.

helpers/images.js
```diff
@@ -49,19 +48,15 @@ export const placeholderRenderOps = {
...
+  const serviceKey = Object.keys(IMAGE_SERVICES).find(s =>
+    IMAGE_SERVICES[s].testUrl(url)
+  );
 
-  // return serviceKey
-  //   ? IMAGE_SERVICES[serviceKey].createUrl(url, renderOpts)
-  //   : url;
+  return serviceKey
+    ? IMAGE_SERVICES[serviceKey].createUrl(url, renderOpts)
+    : url;
```
stephensauceda: Bringing this back

helpers/images.spec.js
```diff
@@ -30,12 +30,6 @@ describe('Image Helper', () => {
       const url = imageUrl(canonicalUrl);
       expect(url).toEqual(serviceUrl);
     });
-
-    it('should use cloudinary fetch for all image URLs', () => {
-      const canonicalUrl = 'https://host/image';
-      const url = imageUrl(canonicalUrl);
-      expect(url).toEqual(createUrl(canonicalUrl));
-    });
   });
```
stephensauceda: not applicable anymore.

lib/imageServices/cloudinary.js
```diff
@@ -8,7 +8,7 @@ import { mapRenderOptions } from './helper';
 export const BASE_URL = 'https://cdn.apartmenttherapy.info/image/upload';
 
 // dynamically create regexp off list of hostnames
-const STORAGE_HOSTS = ['storage.googleapis.com', 's3.amazonaws.com']
+const STORAGE_HOSTS = ['cdn.apartmenttherapy.info']
```
stephensauceda: Honestly not super sure what to do here. Users have the ability to manually paste in an image URL in Flow so there is the possibility one could upload an already-formed `fetch` url (potentially with its own params and things) which would trigger this to return `true`. I know we're not supporting user-provided params at the moment, so maybe it makes sense to check for a `fetch` url in Flow and then just strip it down to the raw storage url?

lib/imageServices/cloudinary.spec.js
```diff
@@ -69,7 +64,7 @@ describe('Cloudinary image service adapter', () => {
       );
     });
 
-    it('should not include any double-slashes between base URL and canonical URL if options is empty', () => {
+    xit('should not include any double-slashes between base URL and canonical URL if options is empty', () => {
```
stephensauceda: I don't _think_ this test is applicable now considering the change in functionality but not altogether sure.

helpers/images.js
stephensauceda: One thing to note, I think `Object.keys` is order-dependent based on how properties are added to the object. So I think that makes `fetch` being a fallback/catch-all for every other URL dependent on the order in which "services" are added to `ImageServices/index.js`. Anything added _after_ cloudinaryFetch to the ImageServices object may cause some false positives in `cloudinaryFetch.testUrl`.

I thought about just making `cloudinaryFetch` check for only the storage urls but that kind of belies its purpose of fetching _all other_ urls.

Might not be a big deal, but just wanted to note.

blech75: ideally, only one `testUrl()` method will return true for a given image URL. i certainly don't want to rely on any order; that would work against the original strategy.

(i haven't looked at the full PR yet, doing that shortly. just wanted to address this comment when i saw it.)

blech75: looking at the code i now see that `testUrl()` for `cloudinaryFetch` is hardcoded to `return true`. i think that needs revising. :-) just need to think about the exact logic....

stephensauceda: Yeah. That's from before. I [had it ](https://github.com/apartmenttherapy/oxygen-lib-ui/pull/235/commits/46ea703fc687907050bcfbd3a9f3b1c2fcdefc3c) looking specifically for the storage urls but changed it back to `true` for everything because I assumed we'd want a fallback that we could control vs simply returning a URL that none of the image services understood (since `fetch` can effectively handle _any_ url we throw to it.)

stephensauceda: At least having `cloudinaryFetch` be the last resort, we can still add our transforms and things to unknown image urls. If we just returned the unknown image, I think there's potential for (layout) breakage if we can't control it's presentation. That problem may be solved elsewhere within all of this though (I'm sure of nothing 😃 )

blech75:
@stephensauceda i took some time to dump some assumptions, thoughts, and questions into an asana task: https://app.asana.com/0/543787180217047/938986206467356/f because, like you...



assuming this is not terribly urgent, let's try to get some clarirty on this next week at retreat so we can all get on the same page.

if it is urgent, well.... we can probably work something out today.

PR #239, opened 2018-12-17
Handle case where item.data is undefined

andrewb:
Courtesy PR. Handle case when `item.data` is undefined. Without this `item.data.placement` _can_ error.


components/PostBundleItem/index.js
```diff
@@ -7,15 +7,14 @@ import ErrorBoundary from '../ErrorBoundary';
...
-
     this.ref = props.includeWrappingDiv ? createRef() : undefined;
   }
 
   render() {
     const { item, customPropsMapper, includeWrappingDiv } = this.props;
-
+    const { data = {} } = item;
     const divClasses = classnames('Post__item', `Post__${item.type}`, {
-      [`Post__${item.data.placement}`]: item.data && item.data.placement
```
blech75: oh jeebus. 🤦‍♂️ check it on one side; ignore it on the other. awesome! (i think that was me 😬)

PR #240, opened 2018-12-18
Henry/transform gif

chenryhen:
Adds a helper to check file (image) paths for a given file extension. Uses that helper to test if a url passed to `Image` is for a gif. If the image is a gif, we use a Cloudinary parameter to transform the gif into an mp4, and then we put it into a <video> tag instead of .

Pull and look at `Image` use cases, primarily images in posts, and images in post galleries. 
`/new-titles-for-everyone-227653?preview=true`
`/snack-cake-recipe-264036`

The animated gifs should be replaced with mp4s

Technically `Image` is used on cover images, sponsor images (like Oxo on the `/recipes` page), and PageHeaders, but none are in use right now.

Just to make sure the video is styled enough (and not in the wrong place with regard to the placeholders or not taking up the proper width)

#### Deployment Notes
There's a corresponding oxygen branch that has more styling, but no functionality additions.

https://app.asana.com/0/543787180217047/948788415866727/f

#### GIF
45afc41355c14f9ba5bbd9bf84731fe5b86bacbe

* [ ] add a reviewer
* [ ] run appropriate tests


components/Image/index.js
```diff
@@ -71,8 +75,19 @@ function Image(props) {
...
+    format: treatAsGif ? 'mp4' : renderOps && renderOps.format
+  };
+
   // CLEANUP: only apply { resizeMode: crop } if maxSize is specified.
-  const finalRenderOps = Object.assign({}, BASE_RENDER_OPS, renderOps);
+  const finalRenderOps = Object.assign(
+    {},
+    BASE_RENDER_OPS,
+    renderOps,
+    gifRenderOps
```
blech75: something like this would make it more succinct, i think:
```suggestion
  const finalRenderOps = Object.assign(
    {},
    BASE_RENDER_OPS,
    renderOps,
    ...(treatImageAsVideo && {
      format: 'mp4'
    })
  );
```

you don't need to worry about an actual ternary expression; the `format` is already in `renderOps`. so just spread the `{ format: 'mp4' }` when you need to.

helpers/images.js
```diff
@@ -78,3 +78,15 @@ export function imageUrl(url, renderOpts = {}) {
...
+
+/**
+ * Checks if an image file path ends with the given extension.
+
+ * @export
+ * @param {String} path image file path
+ * @param {String} extension file extension to look for (no dot)
+ * @returns {Bool}
+ */
+export function matchesFileExtension(path, extension) {
```
blech75: i don't love the name `matchesFileExtension`. but i also don't have any suggestions. :-)

components/Image/index.js
```diff
@@ -89,7 +104,11 @@ function Image(props) {
       : {})
   };
 
-  return (
+  return treatAsGif ? (
```
blech75: (just make sure to adjust this to `treatImageAsVideo` to match the others.)

helpers/images.js
```diff
@@ -78,3 +78,15 @@ export function imageUrl(url, renderOpts = {}) {
...
+  return !!path.toLowerCase().match(new RegExp(`.${extension}$`));
```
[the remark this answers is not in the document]
chenryhen: done. perhaps in 2019 though, I'd like to understand better when the right time to use `!!` is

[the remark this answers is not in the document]
blech75: hehe. this comment was less about `!!` and more about using the appropriate regex method. if you're not capturing/using the matched string, then `.test()` is better/faster.

components/Image/index.js
```diff
@@ -71,8 +75,19 @@ function Image(props) {
     isPlaceholder
   } = props;
 
+  const treatAsGif = matchesFileExtension(url, 'gif') && !isPlaceholder;
```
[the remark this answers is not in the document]
chenryhen: Maybe `convertToVideo`? I chose the name originally because it made more sense to think of it as "treat this image as a gif, and do whatever it is we do for gifs"

components/Image/index.js
chenryhen: yes, classy! I like that much better.

helpers/images.js
chenryhen: maybe `hasFileExtension`? Sounds more like our naming convention at least.

PR #244, opened 2019-01-04
revise imageUrl logic for cloudinary upload/fetch

blech75:
this PR revises how we process image URLs from flow.

first, we adjust the `cloudinary` service to handle the normal/default cloudinary URL, which is "upload". (see https://cloudinary.com/documentation/upload_images). this detection and processing is a bit non-optimal (split/capture using regex), but nothing too crazy.

then, we remove the logic for "finding" the correct image provider for given URL. we just test against `cloudinary` (upload) and if that fails, then fall back to `cloudinaryFetch`.

the net effect is that _any_ non-cloudinary upload URL will be processed through `cloudinaryFetch`. in other words, if the supplied URL is completely external (to another site, for example), then we "fetch" it via cloudinary. i think this is generally a Good Thing since it caches the image on cloudinary (for 7 days) and transforms it to our specs. if this image goes away, however, then we are subject to a 404. previously, we had just returned the original URL, which would still return a 404, but directly hit the external site for every page request and did _not_ transform the image to our specs.

not sure how to test this outside of unit tests since i am not sure what state the flow data is in.


lib/imageServices/cloudinary.js
```diff
@@ -2,12 +2,24 @@ import { baseCreateUrl } from './cloudinary.common';
...
-  // ensure literal dots
-  .map(h => h.replace('.', '\\.'))
-  .join('|');
-export const SERVICE_REGEXP = new RegExp(`${STORAGE_HOSTS}/`);
+// programmatically escape . and /
+const prefix = BASE_API_URL.replace(/([./])/g, '\\$1');
+
+// specifying the image types is a workaround that allows us to craft a regexp
+// that works correctly with extra periods in the file name.
+const imgTypes = ['gif', 'png', 'jpe?g', 'webp'].join('|');
```
blech75: @stephensauceda quick sanity check: am i missing any popular image file extensions here?

stephensauceda: The only "popular" extension I really wonder about is svg. Since it _could_ be used as an `img` `src` I don't know if we want to include it here or what. 

But that also jogged my memory on _all_ file types. [Cloudinary's uploader allows _any_ file type to be uploaded](https://cloudinary.com/documentation/upload_images#uploading_non_image_files_as_raw_files), not just image formats. I'm not sure if another short circuit exists somewhere in the flow through Oxygen to handle that kind of thing. Maybe not a bridge we need to cross here (if it's not already handled elsewhere) but thought I should flag.

blech75: interesting, thanks for the link/reminder. glad we're talking about this! a few thoughts...

* SVG is not a bitmap image, so presumably any transforms will fail. we wouldn't want to attempt to display SVGs via the `imageUrl()`
* same goes for PDFs or any other non-image URLs.

it feels counterproductive to detect or otherwise handle non-image URLs in oxygen; that would ideally be handled on input/upload/selection of an image in flow.

said differently: the image URLs are arriving in oxygen in fields specifically designated for images. we therefore apply `imageUrl()` in good faith, expecting the result to work in whatever context `imageUrl()` is used in. we should never get a non-image URL in an image field.

i can't think of an _existing_ use case for non-image cloudinary URLs being provided to oxygen.

that said, i don't know how we will go about allowing editors (or if we ever have) to upload a generic asset (PDF, XLS, etc.) for linking to in a post. where will that reside? either way, it will not go through `imageUrl()`.

stephensauceda: > SVG is not a bitmap image, so presumably any transforms will fail.

According to [this chart](https://cloudinary.com/documentation/image_transformations#image_format_support) SVG transforms are supported. Given the different ways SVGs can be delivered (even outside of cloudinary) I don't know if this is something we'd _want_ to support though.
> that would ideally be handled on input/upload/selection of an image in flow.
I agree with this. What follows is just further documentation/FYIs. TL;DR I don't know if/how specific we can get in regardless to what we allow to be uploaded and will need to do more poking around with it but it probably isn't a show stopper for this or the Flow side of things either.

I'm seeing some contradictory documentation regarding allowed file types for cloudinary though. [I see](https://cloudinary.com/documentation/upload_widget#upload_widget_options) we can pass a `resourceType` to the actual uploader widget that is rather generic (image, video, raw) which says it _limits_ the allowed files to that "type" leading me to believe you _can't_ upload anything else (which is good, since we can just pass different options into the widget wherever we need; I believe I already support this in the component I made on the Flow side).

The contradiction comes in the upload preset which allows you to define "allowed formats" for all instances of the widget. It says:

> uploaded resources with format not found in the allowed list are converted to the specified format.

which means the file will still be uploaded and attempted to be converted. Weird. The documentation for uploading via their API uses language that sort of straddles the line between these two options. Helpful...

PR #251, opened 2019-01-08
Change post.post id to post.resourceID

seawolff:
- Changes references from `post.postID` to `post.resourceID`.
- See Oxy change here https://github.com/apartmenttherapy/oxygen/pull/1005

- Check SpotIM props
- Check PostBundle props
- Check PostActions props

https://app.asana.com/0/937985426519793/936676642497469/f

* [ ] add a reviewer
* [ ] run appropriate tests


blech75:
@seawolff looks like 83d28f1 sneaked in here. seems like it's a dupe of 16731f6. you must've based this branch off the ingredients header branch by accident (instead of `master`). :-)

anyway, merging this should not be an issue. in the future, you can rebase this branch onto `master` like so:
```
git rebase --onto master feature1-branch feature2-branch
```
where `feature1-branch` is the branch that contained the ingredients header, and `feature2-branch` is the branch that contains the postID work.

seawolff:
@blech75 oops. Yeah that is exactly what I did and somehow missed it. Thanks.

components/PostBundle/index.js
```diff
@@ -35,17 +35,15 @@ function PostBundle(props) {
...
   } = props;
   const {
-    postId,
     canonical,
     title,
     author,
     // CLEANUP: remove createdAt/promotedAt when we remove the fields from the
     // GQL query.
     createdAt,
-    promotedAt,
```
blech75: looks like `promotedAt` and `modifiedAt` were accidentally removed as part of this commit. to keep this branch cleaner, please put them back.

seawolff: @blech75 I can do that but they're not being used. Mind if I commit and add them and then remove them?

blech75: thanks for adding them back in. i didn't want to lose track of them since they were part of a crazy date logic change that is still (somewhat) in progress.

blech75:
@seawolff so i'm wondering about renaming the `SpotIM` component's `postId` prop to `resourceId`. on the surface it seems reasonable (no more `postId` anywhere), but once i started thinking about it, it seemed less desirable (if i understand things correctly... which i may definitely not).

* we currently have comments on posts
* i don't know what our plans are for (spot.im) comments on other types of "standalone" things (recipes, videos, etc.)
* i see that the (former) `postId` prop was used exclusively for setting `data-post-id` on the `<script>` for `spot.im`
* looking at the requests to `spot.im`, it seems that _they_ have some notion of a "post ID" in their comments platform.
* the namespace for `resourceId` and `postId` could very well overlap: we could have a post _and_ a recipe, both with a "resourceId" of 12345. in oder to disambiguate them, we'd need something like a new `resourceType` param/prop). at that point, `spot.im` would need to to know about this "type", too -- its notion of a "post ID" is no longer valid.

what i'm trying to articulate is that it feels like touching the `SpotIM` component is reaching a little "too far into" our components; this change should concern only the post model ad the GQL query that populates the post model. spot.im -- an external service -- is not part of that.

....or maybe i don't have the whole picture, which could very well be the case.

@andrewb thoughts?

seawolff:
I don't really have an opinion on SpotIM. I can change if need be.

andrewb:
FWIW post id is a bit of a misnomer in spot.im. They define it as "any alphanumeric value that uniquely identifies a page" and "a unique identifier for this conversation". So, spot.im markup on other standalone pages will still use `data-post-id`. Our standalone page, which is used for video, has comments enabled already.

That's all a bit of an aside. To get these PRs shipped I propose the following.

Let's keep the prop in the `SpotIM` component as `postId` since internally that component maps it to `data-post-id`, i.e. in `PostBundle` (and the standalone page in oxygen) we do ` <SpotIM postId={data.resourceId} ...etc />`.

seawolff:
@andrewb cool. Amended.

blech75:
@andrewb thanks for the sanity check! you summarized my longwinded explanation nicely. ;-)

good to know about spot.im and their "post id" being somewhat superficial in its naming.

still, note my concern about the simplicity of our post IDs and how other object IDs could collide with them. it's highly unlikely that a JW video ID and a post ID would collide. however, the same could not be said about recipe IDs, which (to my knowledge) follow the same numeric pattern as post IDs. unless we do something like prefix our recipe IDs for spot.im with `recipe-` or something.

anyway.... just mentioning it here to get it out of my head. :-)

PR #252, opened 2019-01-09
Adding recipe description to structured data

seawolff:
Adds recipe description to recipe structured data if present

Use oxygen: https://github.com/apartmenttherapy/oxygen/pull/1014
and oxygen-gql: https://github.com/apartmenttherapy/oxygen-gql/pull/116

Copy post page source into https://search.google.com/structured-data/testing-tool/u/0/

https://app.asana.com/0/786191760618615/969701136036477

#### GIF
image

* [ ] add a reviewer
* [ ] run appropriate tests


components/Recipe/index.js
```diff
@@ -38,6 +39,7 @@ function Recipe(props) {
       <h2 className="Recipe__title" itemProp="name">
         {title}
       </h2>
+      {description && <meta itemProp="description" content={description} />}
```
andrewb: Unfortunately for us `description` is a rich text field in porkchop, so it can contain HTML formatting. For example, on `/recipe-smoky-sweet-potato-amp-black-bean-casserole-239326` I see `<meta itemProp="description" content="&lt;p&gt;A recipe for a hearty vegetarian casserole of sweet potatoes, black beans, and smoked mozzarella that works as both a main or side dish.&lt;/p&gt;"/>`. Note the escaped `<p>` tag.

We'll have to strip tags from `description` otherwise our structured data will include them, literally as `<p>A recipe...</p>`.

Can you find a lightweight way to strip tags from `description`?

seawolff: Sure thing. I can do that.

seawolff: I went ahead and created a helper function to strip the html from the string.

Kind of related: It's a bit annoying because I just realized Google is going to throw a description missing warning even if description meta is present but has an empty string value.

seawolff:
Sure thing. I'll get that out before I merge this.

PR #255, opened 2019-01-14
Henry/new pin

chenryhen:
Refactors the custom `PinItButton` component to use the new Save button documented here:
https://developers.pinterest.com/docs/widgets/save/?

To that end 
- takes out the custom click handler as the new Pinterest script (more about this on the Oxygen side) takes care of what happens on click
- swaps the <button> for an <a> per the docs
- moves `PinItButton` out of `LazyLoad` so that the Pinterest script can process the button on page load.

You will need the companion `oxygen` PR: https://github.com/apartmenttherapy/oxygen/pull/1025

Pull and go to any post page. You should be able to click the Pin button on standalone image or gallery image, and you'll see a pop-up that lets you add that image to a Pinterest board.

Unrelated note: clicking on the Pinterest button on a gallery slide expands the gallery. I will take care of that in a separate PR

#### Deployment Notes

https://app.asana.com/0/937985426519793/919759511279010/f

#### GIF
pin_placement

* [ ] add a reviewer
* [ ] run appropriate tests


components/Image/index.js
```diff
@@ -101,7 +101,8 @@ function Image(props) {
       sizes:
         sourceSizes && sourceSizes.length > 0
           ? buildSizes(sourceSizes)
-          : undefined
+          : undefined,
+      'data-pin-id': url
```
blech75: @chenryhen if i'm reading the docs correctly, it seems this should go on the link (`a.PinItButton`).

chenryhen: perhaps I will put it on both because honestly their documentation leaves a lot to be desired. `data-pin-id` and a few others have a note `This attribute is applied individually to images on your page.` I'm _assuming_ that means it does not go on the button itself. I found an example somewhere else that has it on the  too. 

aha! https://business.pinterest.com/en/blog/pin-it-button-technical-tune-up-5-tips-to-make-sharing-from-your-site-better

blech75: ok, after thinking about this for a few min, it turns out you're right. lol

the attrs are useful on the individual images for the use case of pinning the page in the `buttonBookmark` mode -- something we're not directly using.

presumably, the `buttonBookmark` mode is functionally identical to when one uses the browser extension to pin a page: you're presented with all of the images found on the page. in this case, we need the data attrs on the individual image elements to provide the correct context.

hope that makes sense!

blech75:
@chenryhen i'm closing this out. superseded by apartmenttherapy/oxygen#1166

PR #266, opened 2019-01-23
Keaton/gallery

keaton-taylor:
Layout changes for galleries adding in a legacy-adjacent mixin to make the swap-over less complex. 

https://app.asana.com/0/937985426519793/998099379245961/f
https://github.com/apartmenttherapy/oxygen/pull/1050


components/Gallery/index.scss
```diff
@@ -149,6 +152,86 @@ Classes that jump between different variants depending on where they are on the
...
+
+  .Gallery__slide {
+    display: grid;
+    grid-area: slides;
+    grid-template-areas:
+      'slide slide'
+      'number caption';
+    grid-gap: 1em;
+    grid-template-columns: 4.75rem 1fr;
+    grid-template-rows: auto 40px;
```
chenryhen: screenshot 2019-01-23 15 43 23

You've got some messy overlapping here, if some of the captions are long. I would make `grid-template-rows: auto auto` There will be some up and down jank as the caption changes length, but that's an acceptable trade-off.

keaton-taylor: Whoa. That didn't preview that way for me. Ugh...

PR #269, opened 2019-01-24
return null if there isn't anything to map

chenryhen:
Quick fix for `0` showing up in Recipes. As near as I could tell, the `<Fragment>` from `RecipeSupplies` was still rendering the zero even when the `groups` prop was empty. Now it explicitly return `null` to prevent anything from showing up.

@blech75 @andrewb whoever can nab this first, that would be great.


blech75:
@chenryhen sorry, this was my fault. 😬  i'll look at it now.

blech75:
@chenryhen so i completely changed the way things were tested to fill some gaps. i also ensure that groups and items worked the same way. whew.

PR #284, opened 2019-02-07
Large ads go wider than 300px

seawolff:
Courtesy PR to make it so Ads can't go wider than 300px

Part of https://app.asana.com/0/937985426519793/1104702281788896/f


components/Ad/index.critical.scss
```diff
@@ -3,6 +3,7 @@
   justify-content: center;
   opacity: 1;
   transition: opacity 1s;
+  max-width: 300px; // I checked and this is a known Ad size. Prevents Ads fitting width of Rail container
```
andrewb: @seawolff we don't want to apply this to all ads since we have a lot of different sizes, e.g. see in post:

screen shot 2019-02-07 at 7 25 57 pm

Note how it's now misaligned because of the max width on the container.

I think [postAds.scss](https://github.com/apartmenttherapy/oxygen/blob/master/components/PostBundle/postAds.scss) would be better, as well as a specific rule targeting the rail ad, e.g. `.PostBundle--with-rail .Rail .Ad` (or something like that).

PR #290, opened 2019-02-13
Keaton/isvideo icon

keaton-taylor:
Adding this isVideo tag conditional


keaton-taylor:
@chenryhen looking better. Only one icon.

PR #303, opened 2019-02-20
Reduce lodash deps

andrewb:
Courtesy PR. This removes a few lodash dependencies, namely `pick`, `uniq` and `camelcase` (this one was included due to another dependency).

See individual commits for details but tl;dr is `pick` is replaced by using `reduce`, `Set` can replace `uniq` and instead of camel casing I've switched to using the slug (this works with the existing sponsors).

This doesn't save us much, just a few kb gzipped. But less is better and so are less deps. It also lets me procrastinate about what to do with larger deps like moment.


components/SpreadBundle/helper.js
```diff
@@ -94,15 +92,18 @@ export function createSetBundleItemProps(pictureSpecs, exceptionPictureSpecs) {
     if (type === 'teaser') {
       // thin out the passed data based on the variant so the teaser does not
       // render unwanted props.
-      baseData = pick(
-        baseData,
-        VARIANT_TEASER_PROPS[variant] || DEFAULT_TEASER_PROPS
-      );
+      baseData = Object.entries(baseData).reduce((p, [key, value]) => {
```
blech75: @andrewb is `Object.entries()` transpiled/polyfilled by babel? (how do i check that sort of thing?)

andrewb: @blech75 yep it is, using https://babeljs.io/docs/en/babel-polyfill and https://github.com/zloirock/core-js (which has a handy list).

PR #306, opened 2019-02-21
Add width to mp4 to fix broken iOS playback

andrewb:
Courtesy PR. Add width to mp4 to fix video playback issue in iOS.


components/Image/index.js
```diff
@@ -85,7 +85,8 @@ function Image(props) {
     BASE_RENDER_OPS,
     renderOps,
     convertToVideo && {
-      format: 'mp4'
+      format: 'mp4',
+      width: 720
```
blech75: @andrewb i know this was just a quickie to get it working. thank for figuring it out!

stepping back, though, a few comments/thoughts (@chenryhen please chime in!):

* `720` feels a bit 'magic'. what about using `POST_COL_WIDTH`, which is `730`?
* or should we instead increase the `width` to `POST_COL_WIDTH * 2` to account for hi-dpi? just anticipating IQ feedback re: the animated version appearing different/crappier than the other retina-sized images. :-)
* in any case, hardcoding the `width` here obviously makes it so that the video is not optimized for the current viewport width. this probably not a huge deal, and i imagine we might not want to go down the road of responsive videos -- is there such a road? (the `width` is not usually/yet specified here because images get `srcSet` populated with the various sizes. is there an equivalent declarative approach for videos?)

just for kicks, i looked at the filesize for the `bona-MAKE-YOUR-BED.gif` as an MP4 at different rendered widths:

* [320px](https://cdn.apartmenttherapy.info/image/fetch/f_mp4,q_auto:eco,c_fill,g_auto,w_320/https://storage.googleapis.com/gen-atmedia/2/2019/02/bona-MAKE-YOUR-BED.gif): 30,330 bytes
* [480px](https://cdn.apartmenttherapy.info/image/fetch/f_mp4,q_auto:eco,c_fill,g_auto,w_480/https://storage.googleapis.com/gen-atmedia/2/2019/02/bona-MAKE-YOUR-BED.gif): 55,630 bytes
* [720px](https://cdn.apartmenttherapy.info/image/fetch/f_mp4,q_auto:eco,c_fill,g_auto,w_720/https://storage.googleapis.com/gen-atmedia/2/2019/02/bona-MAKE-YOUR-BED.gif): 86,303 bytes
* [1280px](https://cdn.apartmenttherapy.info/image/fetch/f_mp4,q_auto:eco,c_fill,g_auto,w_1280/https://storage.googleapis.com/gen-atmedia/2/2019/02/bona-MAKE-YOUR-BED.gif): 192,837 bytes
* [1440px](https://cdn.apartmenttherapy.info/image/fetch/f_mp4,q_auto:eco,c_fill,g_auto,w_1440/https://storage.googleapis.com/gen-atmedia/2/2019/02/bona-MAKE-YOUR-BED.gif): 221,302 bytes

FWIW, i output the same as GIFs, and they're running a consistent 2x the filesize of the MP4.

andrewb: Yeah, it was somewhat arbitrary and can change. I went with 720 since it's a common size for video. Of course these are different contexts, so they don't have to match. We should go as small as we can get away with imo.

PR #307, opened 2019-02-22
compile postinstall

stephensauceda:
Admittedly, I'm not 100% sure why we consume lib-ui the way we do so if I'm missing something, please inform me. But currently, we have to jump through some hoops to compile this library when it lives in `node_modules` which, for some reason unbeknownst to me, broke when we tried to upgrade the workflow app to Next 8. I'll add some details below if you're interested but they are not exactly relevant to this PR, just know that the non-standard method of having to compile this one package in `node_modules` seemed to be the problem.

<details>
<summary>Next 8 problem</summary>

Basically, for some reason, Webpack/Babel was having trouble finding relative files within lib-ui when it tried to compile it. I'm not entirely sure what caused it because there didn't seem to be a big Webpack or Babel update included with Next 8 (though I of course could have missed something). My guess is, they changed some settings in _their_ Babel loader config since, near as I can tell, we were just using it as the loader for lib-ui. I haven't messed with Webpack configs in like two years and I just c/p the lib-ui stuff from Oxygen when I brought lib-ui into Workflow so, 🤷‍♂️. 

</details>

Trying to fix the webpack config seemed like treating the symptom rather than the illness, so I went this route instead. Compiling lib-ui postinstall means we don't have to selectively pick it out of `node_modules` for compilation and generally makes life easier on the consumption side of things. It shouldn't have any affect on any other app (Oxygen) that consumes this library if you already have it set up but you should just be able to change your `import`s from `oxygen-lib-ui/components/<whatever>` to `oxygen-lib-ui/dist/components/<whatever>` to consume the pre-built components. That's all I had to do to get it [working in workflow](https://github.com/apartmenttherapy/gen-workflow/compare/stephen/next8) anyway (what's on [playground](https://workflow-dot-playground-194317.appspot.com/) right now is running lib-ui in this manner).


package.json
```diff
@@ -12,7 +12,9 @@
     "storybook:build": "build-storybook -c .storybook -o www",
     "dev": "yarn run test:json && yarn run storybook:start",
     "deploy": "yarn run test:json && yarn run storybook:build && gcloud app deploy --project=gen-prod --no-promote",
-    "deploy:promote": "yarn run test:json && yarn run storybook:build && gcloud app deploy --project=gen-prod --promote --stop-previous-version"
+    "deploy:promote": "yarn run test:json && yarn run storybook:build && gcloud app deploy --project=gen-prod --promote --stop-previous-version",
+    "build": "babel . -d dist --ignore '*.spec.js','*.story.js','node_modules'",
+    "postinstall": "yarn build"
```
stephensauceda: I looked at using the `prepare` lifecycle script instead of `postinstall` but it didn't appear to be running when the package was installed for some reason. Six of one, half dozen of the other in my mind (at least for this use case).

package.json
```diff
@@ -54,6 +52,12 @@
   },
   "dependencies": {
     "@atmedia/react-gpt": "^2.1.1",
+    "babel-cli": "^6.26.0",
+    "babel-core": "^6.26.0",
+    "babel-loader": "^7.1.4",
+    "babel-plugin-transform-object-rest-spread": "^6.26.0",
+    "babel-preset-env": "^1.7.0",
+    "babel-preset-react": "^6.24.1",
```
stephensauceda: just moving this stuff to make sure it's all installed and can be used to compile when the package gets installed.

blech75:
one quick comment and then i'll try to dive in deeper a little later today...

IIRC, our rationale/motivation for this setup for purely for its interplay with oxygen.

first things that occur to me:

* ease of developer experience: components in lib-ui act the same as oxygen; they're just in a different package.
* duplication of webpack configs across  (e.g. critical/amp CSS, etc.)

plus other stuff, i'm sure.

you certainly raise discussion-worthy points; thanks for bringing it up. now that lib-ui is consumed by more than just oxy we have to  consider the ramifications of our earlier decisions!

also: we're gonna take a look at upgrading oxygen to next8, so we may well hit the same issue you did. :-)

blech75:
@stephensauceda good news (i think).

we just tried upgrading oxygen to next 8 and ran into the same issue you did in workflow. in reviewing the transpiling config, i revisited the old PR and saw there is now a plugin to handle this: `next-transpile-modules`. turns out it was forked and recently updated for next 8!

here is what i did to get next 7 --> 8 working on oxygen: https://github.com/apartmenttherapy/oxygen/commit/9d206ef40edccdd9e79ea817f2b6236ee206bfa5

give it a try on workflow and lemme know if/how it works for you!

seawolff:
@blech75 looks like they have 2.0 live with the Next 8 support. I don't think we will need to fork. https://github.com/martpie/next-transpile-modules/releases/tag/2.0.0

blech75:
@seawolff ah, so it seems that martpie's fork _is_ what is published as `next-transpile-modules` on npm: https://www.npmjs.com/package/next-transpile-modules 🎉 (since i noticed his repo was a fork, i glossed over the fact that the package he forked from was published as `@weco/next-plugin-transpile-modules` 🤦‍♂️ )

that means we can change the reference from his github repo:
```
"next-transpile-modules": "martpie/next-transpile-modules#2.0.0",
```
to the actual package on npm:
```
"next-transpile-modules": "^2.0.0",
```
ah... so much more formal!

PR #310, opened 2019-02-22
Use data.kicker to generate kicker value

andrewb:
This updates the spread bundle to use the new bundle `data.kicker` value to generate the correct kicker for each teaser in a bundle. The value can be `timestamp` (current default), `tag` or `category`. This value applies to all teasers in a bundle.

See https://app.asana.com/0/937985426519793/1110392282723399 for an explanation of how this should work.

The teaser's kicker value gets overridden with the new value. As a follow up we can get Flow to remove `kicker` from the teaser data so there is no confusion where this value comes from.

The kicker value gets added before the teaser fields are filtered. This means if a teaser variant does not support kicker, the value will still be removed.

Since `timestamp` is now the default the kicker will show the relative date in the first module of the homepage.

See also https://github.com/apartmenttherapy/oxygen-gql/pull/132.

To test you'll need https://github.com/apartmenttherapy/oxygen/compare/andrew/kicker. Go to...


chenryhen:
@andrewb I should've checked the card more thoroughly, but I'm hoping we can adjust this behavior a little on our end. I understand that it's easier for Editorial to think of them as interchangeable, but I feel like it's better if we keep the timestamp in the `promotedAt` prop (and probably rename it to `timestamp`)  and place the category/alternative in `kicker`. Let me know if you want to chat about this.

PR #313, opened 2019-02-26
Keaton/video icon

keaton-taylor:
Makes strategic, presentational changes to existing teaser mixins.


components/Teaser/variantMixins.scss
```diff
@@ -200,11 +245,58 @@
     }
 
     .Teaser__body {
-      flex: 1 0 0px;
+      flex: 1 0 0;
```
chenryhen: was this a simplification or did it change the behavior. I think a unit-less flex basis breaks on older browsers. https://github.com/philipwalton/flexbugs#flexbug-4

components/Teaser/variantMixins.scss
```diff
@@ -200,11 +245,58 @@
...
   }
+
+  @supports (display: grid) {
+    display: grid;
+    grid-template: 'picture body' / 1fr 1fr;
+    grid-gap: 1em;
+
+    .Teaser__body {
+      grid-area: body;
+      flex: none;
```
chenryhen: you don't need this. Once it's a grid item, the `flex` rule is ignored.

components/Teaser/variantMixins.scss
```diff
@@ -200,11 +245,58 @@
...
     .Teaser__body {
-      flex: 1 0 0px;
+      flex: 1 0 0;
       max-width: none;
       min-width: 0;
     }
   }
+
+  @supports (display: grid) {
+    display: grid;
```
chenryhen: I would move all of this general Teaser styling into `&.Teaser--video` because we don't need to switch it to grid context until it's a `Teaser--video`

components/Teaser/variantMixins.scss
```diff
@@ -200,11 +245,58 @@
...
+        top: auto;
+        left: auto;
+        // Put it in the same grid area as the image
+        display: grid;
+        grid-area: picture;
+        align-self: center;
+        justify-self: center;
+      }
+    }
+  }
```
chenryhen: We'll need to adjust the grid columns for places where the `$isFlexible` parameter of the mixin is set to true. Hit me up and I can talk though this if it's not clear.

components/Teaser/variantMixins.scss
keaton-taylor: Moved

components/Teaser/variantMixins.scss
keaton-taylor: fixed

components/Teaser/variantMixins.scss
keaton-taylor: fixed

PR #314, opened 2019-02-26
Henry/post sponsor

chenryhen:
-  Changes `BundleSponsor` so it assumes Workflow/Wordpress users won't have the foresight to add a trailing space at the end of the messaging string
-  Updates PostBundle to expect a `sponsor` in its data.
-  Replaces the old cake tag with `BundleSponsor` 

You'll need to run this `oxygen-gql` branch locally but still use the production `FLOW_URI`: https://github.com/apartmenttherapy/oxygen-gql/pull/133

as well as the companion `oxygen` branch: TBD

#### Deployment Notes

* [ ] add a reviewer
* [ ] run appropriate tests


components/BundleSponsor/index.js
```diff
@@ -25,7 +25,12 @@ function BundleSponsor({ cssHook, imageUrl, messaging, name }) {
 
   return (
     <div className={baseClass}>
-      {messaging && <span className={nodeClasses.messaging}>{messaging}</span>}
+      {messaging && (
+        <span className={nodeClasses.messaging}>
+          {messaging}
+          &nbsp;
```
blech75: hmm... this `&nbsp;` is causing test failures:
```
 FAIL  components/BundleSponsor/index.spec.js
  ● <BundleSponsor /> › props › should show props.messaging

    expect(received).toEqual(expected)
    
    Expected value to equal:
      "qwerty"
    Received:
      "qwerty "
```
since it's a non-breaking space char, testing for equality is slightly awkward because you have to insert an actual nbsp unicode char into the string (which looks identical to a normal space):
```
expect(messagingEl.text()).toEqual(`${messaging} `);
```
that's one in there! copy/paste, or do it yourself by searching for 'no-break space' in your char palette:

screen shot 2019-02-27 at 10 04 42 am

components/PostBundle/index.js
```diff
@@ -126,7 +128,18 @@ function PostBundle(props) {
           dangerouslySetInnerHTML={{ __html: title }}
         />
 
-        {hasTagSlug(tags, SPONSORED_TAG) && <div className="Post__cake-tag" />}
+        {hasTagSlug(tags, SPONSORED_TAG) && (
+          <div className="Post__cake-tag">
+            {sponsor &&
```
blech75: this is neither here nor there, but i might refactor `BundleSponsor` to use a pattern where you can just do `{sponsor && <BundleSponsor ... />}` which has the advantage of keeping the conditionals as simple as possible. no `BundleSponsor` props are required, so there's no harm in passing `undefined`/`null` values from an incomplete `sponsor` object.

of course, you'd then have to shift the logic for conditionally outputting the surrounding `<div>` to the `<BundleSponsor>` component itself. worth it, IMHO, because that would make it simpler to use.

i just mention this because of [the other use](https://github.com/apartmenttherapy/oxygen-lib-ui/blob/master/components/SpreadBundle/index.js#L94-L102) of `<BundleSponsor>` in which looks like `{sponsor && (sponsor.imageUrl || sponsor.messaging || sponsor.name) && (<BundleSponsor ... />)}`

components/BundleSponsor/index.js
chenryhen: oh weird. I guess my test was just lagging behind the changes. I'll fix that

blech75:
@chenryhen oh, and nice work on the tests!

components/PostBundle/index.js
chenryhen: I _think_ I can do this now. Most of it is already conditional in `BundleSponsor`

chenryhen:
I guess if it's been approved, you can't submit for re-review. @blech75 do you mind taking another look since I moved some logic around? It's working for me locally, and I fixed the test too.

components/BundleSponsor/index.js
```diff
@@ -23,33 +23,31 @@ function BundleSponsor({ cssHook, imageUrl, messaging, name }) {
     }
   ];
 
-  return (
+  return name || imageUrl ? (
```
blech75: one change i'd make for readability here would be to bail out early with `null`. the front-loaded ternary buries the `null`.

something like this:
```
function BundleSponsor({ cssHook, imageUrl, messaging, name }) {
  if (!name && !imageUrl) {
    return null;
  }
  // ...
```

that effectively makes `name` or `imageUrl` required and avoids executing any further logic.

PR #322, opened 2019-03-06
Henry/storybook updates

chenryhen:
I'm trying to get Storybook back to being a useful resource, and the first step was to clean up instances where the stories fell behind the updates to the components themselves. A bunch were tripping over prop changes. The most convoluted one was the `EmailSignup` -> `EmailSignupForm` -> `EmailSignupConfirmation` chain because they didn't actually use prop shapes that matched the `SOCIAL` constants from Oxygen. And then there was an inconsistency with what `ActionIcon` expected, so I ironed that out. 

So, you should be able to see non-errors for all of the Storybook components now. Also the sites should still work. The main thing to check is `/email` because those were the only ones where I really changed anything in the components themselves.

Edit: I lied. Please sanity check the `I Made This` in recipes as well.


chenryhen:
@blech75 I completely forgot about this. I've cleared all the errors except `ErrorBoundary` (which I think is intentional) and the `SpreadBundle` errors about duplicate keys, which I still can't track down.

blech75:
@chenryhen you don't have to worry about the `key` errors; that appears to be from within the storybook react app itself (i see references to glamorous and tables).

PR #328, opened 2019-03-13
Henry/spotim ignore

chenryhen:
Add a prop to SpotIm to conditionally add a data attribute to the Spot.IM launcher. This attribute should prevent the post that the launcher is on from being used in Spot.IM recirc. Per advice from Spot.IM here https://app.asana.com/0/937985426519793/1113193571082768/f

Still working on a way to validate that this is actually working.


chenryhen:
Updated `PostBundle/index.js` to look for an action tag on the post. As always, naming is hard, so any preferences there are welcome. Copying this over from the `oxygen` PR for reference:

> You need the companion PR ^. Restart Oxygen in K.
> 
> load /paula-deen-meatloaf-266036. Check that the element with data-spotim-module="spotim-launcher" has the attribute /data-ignore-host-url="true". No other post on the site should have this.
> 
> Comments should work regardless.

I've added the tag `action:block from spotim` to only that post for now ^

components/SpotIM/index.js
```diff
@@ -149,15 +152,17 @@ class SpotIM extends Component {
...
 };
 
 SpotIM.propTypes = {
   spotId: string.isRequired,
   postId: string.isRequired,
   canonical: string.isRequired,
   sso: func,
-  isUserSignedIn: bool
+  isUserSignedIn: bool,
+  ignoreResource: bool
```
andrewb: Tiniest nit, but maybe `ignoreHostUrl`, or `blockFromRecirc`?

chenryhen: no prob

PR #329, opened 2019-03-13
update ItemBundle to pay attention to `data`

chenryhen:
There are related oxygen and oxygen-gql branches to go with this. I'll note them in the oxygen PR. 

This adds the `data` prop to `ItemBundle`. The data is being sent through already, we just need to grab it for the component.

oops. Forgot to mention that it also uses the color to set a `theme--*` classname that sets a CSS custom property in oxygen.


blech75:
neither here nor there, i feel like we should rename `ItemBundle` to something like `DefaultBundle`, `BasicBundle`, or `SimpleBundle`. it took me a few beats to remember wtf this component does. i had to read your excellent comment in the component to remind me :-)

PR #332, opened 2019-03-15
Henry/sponsored email

chenryhen:
Added the ability to add a sponsor to the `EmailSign` component. In hindsight now, `BundleSponsor` is probably not the best name anymore, but trying to limit the scope of these changes.


components/EmailSignup/index.js
```diff
@@ -81,17 +83,22 @@ EmailSignup.propTypes = {
   listDataError: shape({ msg: string }),
   listKey: string.isRequired,
   // not required because it gets applied via cloneElement and causes prop type
-  // check errors on load
```
blech75: the `not required ... check errors on load` was a single comment that applied to the `subscribeUserToList` prop. it got disassociated from that prop during your alphabetization :-)

chenryhen: 😬

components/EmailSignup/index.js
```diff
@@ -81,17 +83,22 @@ EmailSignup.propTypes = {
...
-  variant: string,
+  // check errors on load,
   socialLinks: arrayOf(
     shape({
       href: string,
       label: string
     })
-  ).isRequired
+  ).isRequired,
+  sponsor: shape({
```
blech75: maybe use `shape(EmailSignupForm.propTypes.sponsor)`?

components/EmailSignupForm/index.js
```diff
@@ -99,6 +101,7 @@ class EmailSignupForm extends Component {
           <section className="EmailSignup__body">
             {title && <h2 className="EmailSignup__title">{title}</h2>}
             {description && <p className="EmailSignup__desc">{description}</p>}
+            {sponsor && <BundleSponsor {...sponsor} />}
```
blech75: per discussion, spread `cssHook` here and remove the default value below. (props are shallowly merged, so this avoids any chance of `cssHook` getting _unset_.

components/EmailSignupForm/index.js
```diff
@@ -168,6 +171,7 @@ EmailSignupForm.propTypes = {
     })
   ).isRequired,
   optInEvent: string,
+  sponsor: shape(BundleSponsor.propTypes),
```
blech75: and then define the shape for `sponsor` (minus `cssHook`) here.

PR #336, opened 2019-03-18
Update to use name/slug for terms

andrewb:
Companion to https://github.com/apartmenttherapy/oxygen/pull/1241. This removes `toSlug` since this is now handled in `gql` and reworks cats and tags to use `name` and `slug`. See https://github.com/apartmenttherapy/oxygen/pull/1241 for details.


components/BundleSponsor/index.js
```diff
@@ -14,8 +13,9 @@ function BundleSponsor({ cssHook, imageUrl, messaging, name }) {
 
   // We use "Cake" to avoid adblock class/id blacklists
   const baseClass = `${cssHook}__cake`;
+  const slug = name ? name.toLowerCase().replace(/\s/gi, '-') : '';
```
blech75: is this robust enough? @chenryhen ?

blech75: chatted with @chenryhen. this is fine for now, but just noting that it does not yield the same results as:
```
        export const toSlug = term => {
	  const options = {
	    replacement: REPLACEMENT,
	    lower: true,
	    remove: /[*+~#=^;.,()[\]"!@]/g
	  };
	  // Slugify term (and also replace `:` and `'` with `-`)
	  return slugify(term, options).replace(/[:']/g, REPLACEMENT);
	};
```

helpers/terms.js
```diff
@@ -0,0 +1,17 @@
+export const hasTermSlug = (terms = [], slug) =>
```
blech75: my initial impression is that the name `hasTermSlug` feels a bit awkward, but that's probably because i haven't touched this code yet. `hasTerm()` feels more natural, but i know that has other tradeoffs. we can revisit later. just mentioning my 2¢.

PR #341, opened 2019-03-20
Henry/picture video poc

chenryhen:
So I originally thought <source> in <picture> supported the `media` attribute, but it turns out it explicitly does _not_ ha. So that made swapping out <picture> gifs for <video> easier, but less responsive. I think I've got a match of code and styling that should support _most_ uses of an animated gif in a teaser.

In summary, I swap out the entire <picture> with all of its sources for one <video> element, similarly to how we're doing it for `Image`. I'm assuming the width of the image for the widest <source> media query is as big as the video will ever need to be.

There is some funky styling (especially in `H` Porthole) on Safari, that I'm trying to iron out, but I've looked at it in a few rectangular teasers and it seems to be working as expected. All thought are welcome. There's currently a K stream `moderated:test-stream-to-exclude-tags-214810001` that has a teaser with an animated gif pinned to the first teaser, if you want to play around or see it on `/collection/bundle-samples`

On a slightly embarrassing note, for simple gifs, sometimes compression leads to a smaller size than the mp4 anyway. But I still think the mp4 conversion is better to cover all cases.

@andrewb @blech75 how much further do you think this update needs to go before we feel ok putting it in production?


blech75:
a few things:

1) tests are currently failing because you're assuming that `sourceSets` will always contain something. the default sanity check for the spec does not pass in `sourceSets`. that might have been an oversight on my part when i refactored things. (sorry!) if we change `sourceSets` to required and pass in a single source set, we're golden.


2) in any case, though, the `videoWidth` logic could be more clearly written by using an intermediate var:
    ```js
    // Making an assumptions that the largest widths are used
    // at the largest media query.
    const largestSize = sourceSets[0].layoutWidth;
    const videoWidth =
      largestSize.units === 'px'
        ? // if the units are pixels, we just use the value
          largestSize.value
        : // if the units are not pixels, we assume they are vw
          // so we use the units as a percentage of 1024
          // which is when we stop using a fluid responsive layout.
          (1024 * largestSize.units) / 100;
    ```
    this highlights something else. our proptype checks will flag any bad data, but we could still have runtime error attempting to 1) either get `layoutWidth` on an undefined value (when `sourceSets[0]` does not exist) and 2) getting `units` or `value` on an undefined (if `layoutWidth` doesn't exist).

3) i dislike now we're co-mingling concerns of normal pictures and videos in the `Picture` component. as you have it, the code path does computes the `sources` even if we're displaying a `<video>`. this could use a refactor to split them up, but we might be able to get away with separate returns instead of a single return with a ternary. will have something soon to show you if you'd like.

PR #344, opened 2019-03-26
add teaser kicker logic

blech75:
this PR implements logic to set or prune key teaser props (`kicker`, `promotedAt`) based on the value of the `kicker` prop in the bundle.

see https://app.asana.com/0/0/1113609380452917/f

companion PR to https://github.com/apartmenttherapy/oxygen/pull/1275


components/SpreadBundle/helper.js
```diff
@@ -9,6 +11,7 @@ const DEFAULT_TEASER_PROPS = [
   'imageUrl',
   'url',
   'tags',
+  'categories',
```
blech75: i'm passing `categories` to teasers now. not that we necessarily need them. thoughts?

components/SpreadBundle/helper.js
```diff
@@ -119,6 +128,21 @@ export function createSetBundleItemProps(pictureSpecs, exceptionPictureSpecs) {
           variant
         };
       }
+
+      const category =
+        data.categories && data.categories.length > 0
+          ? data.categories[0].name
+          : '';
```
blech75: i've noticed that some teasers/posts have categories set to `null`, specifically this one:
```json
{
  "data": {
    "author": null,
    "headline": "Read All the Latest from Kitchn",
    "imageUrl": "https://s3.amazonaws.com/pixtruder/original_images/97df5012a6b071e26a31b9d2de06861b25898c1e",
    "tags": null,
    "promotedAt": 1552078800,
    "imageId": null,
    "url": "https://www.thekitchn.com/latest",
    "dek": "Click here for a full feed of all our latest posts. ",
    "kicker": null,
    "categories": null,
    "cta": null
  },
  "type": "teaser",
  "id": "2e927e4f564fd046ef16469ca3ac7e3c35e9ca49"
}
```

i think for posts, though, we can assume `categories` is always present.

components/SpreadBundle/index.js
```diff
@@ -80,10 +80,11 @@ function SpreadBundle(props) {
   // create the method that assigns props to the bundle items. this needs to be
   // a partially-applied function so we can handle different props
   // (basePictureSpecs, exceptionPictureSpecs) for each variant.
-  const setBundleItemProps = createSetBundleItemProps(
-    bundlePictureSpecs.base,
-    exceptionPictureSpecs
-  );
+  const setBundleItemProps = createSetBundleItemProps({
```
blech75: things are getting awkward here with the addition of `bundleData`....

components/SpreadBundle/helper.js
chenryhen: ah shoot. I noticed this as well. Video permalink teasers might also have the same issue. For visual consistency, I like the idea of filling the `kicker` with _something_ though. For videos, that's pretty easy, we could just use "Video". For everything else, I dunno…

components/SpreadBundle/helper.js
chenryhen: IMO, if `Teaser` isn't setting the kicker or using the categories at all I don't think there's any need to pass them, right?

components/SpreadBundle/helper.js
```diff
@@ -119,6 +128,21 @@ export function createSetBundleItemProps(pictureSpecs, exceptionPictureSpecs) {
...
+
+      // set teaser kicker based on what bundle specifies. bundle data values
+      // are set in flow's UI code.
+      if (bundleData.kicker === 'timestamp') {
+        moreData.kicker = getRelativeDate(data.promotedAt);
```
chenryhen: Ok, so here is where I have questions! :D per our Slack convo, can this be
```
if (bundleData.kicker === 'timestamp') {
        moreData.timestamp = getRelativeDate(data.promotedAt);
```
So if the Flow concept of kicker is set to `timestamp` we fill the Teaser's `timestamp` and not its `kicker` (leave it `undefined`, and then the Teaser won't render the kicker)

Likewise, in the next two cases, we pass in a `moreData.kicker` but we leave timestamp `undefined`. Again, so the timestamp doesn't render. Is that crazy?

components/SpreadBundle/helper.js
blech75: you are correct, good sir. removing....

components/SpreadBundle/helper.js
blech75: ugh. i need to think about this. this is getting more complicated....

components/SpreadBundle/helper.js
blech75: let's chat later about this. i am thoroughly confused, possibly due to lack of nourishment.

chenryhen: prop is actually called `promotedAt`

components/SpreadBundle/helper.js
blech75: @chenryhen can you clarify what you mean by "Video permalink teasers"?

chenryhen: @blech75 we're going to start promoting teasers to the standalone video pages (e.g. https://apartmenttherapy.com/pack-a-suitcase-with-marie-kondo-video:okMNIgOr) in bundles. These don't have any `kicker` set on them in the data. The top three teasers are for standalone videos at apartmenttherapy.com/collection/video

PR #356, opened 2019-04-10
New SpotIM Auth integration

seawolff:
This PR uses https://github.com/apartmenttherapy/oxygen/pull/1300

It replaces a fetch to the old community service for sso and replaces it with a mutation call to the user profile service.


seawolff:
@chenryhen Odd CSS question for you. Seeing this strange CSS rule not being supplied to the SpotIM component. Any idea?

_My branch_
Screenshot 2019-04-10 16 48 01
_Production_
Screenshot 2019-04-10 16 47 49

chenryhen:
Ah @seawolff that login link is supplied on our end. Because we had our own auth, we had to put our own link above the SpotIM module. It gets styled contextually, like here: https://github.com/apartmenttherapy/oxygen/blob/master/components/PostBundle/postComments.scss#L35

seawolff:
For reference this was fixed with https://github.com/apartmenttherapy/oxygen/pull/1300/commits/302f6ecefe6c1d690e598f9b86671169b93da6f3

blech75:
@seawolff please resolve conflicts with master after you merge #368

components/PostBundle/index.js
```diff
@@ -330,7 +335,22 @@ PostBundle.propTypes = {
 
   extraArticleClasses: arrayOf(string),
   itemPropMapperAdditions: func,
-  pagePath: string.isRequired
+  pagePath: string.isRequired,
+
+  // user is authenticated in auth0
+  user: shape({
```
blech75: i would avoid re-defining the `user` props here. since you're not directly using them, i'd suggest doing something like `shape(SpotIM.propTypes.user)`

components/PostBundle/index.js
```diff
@@ -5,7 +5,8 @@ import {
   func,
   shape,
   string,
-  number
+  number,
+  objectOf
```
blech75: `objectOf` appears unused

components/SpotIM/index.js
```diff
@@ -94,13 +101,22 @@ class SpotIM extends Component {
...
+      })
+    )
       .then(() => {
         this.setState({
           authorized: true
         });
       })
-      .catch(() => {
+      .catch(error => {
+        console.log(error);
```
blech75: @seawolff did you mean to leave this `console.log()` in?

seawolff: Yes. It's just for debugging.

components/PostBundle/index.js
```diff
@@ -205,6 +208,8 @@ function PostBundle(props) {
           <div id={`comments-${resourceId}`} className="PostComments">
             <LazyLoad height={200}>
               <SpotIM
+                createSpotIMIntegration={spotimIntegration}
+                user={user}
                 spotId={spotIM.id}
                 sso={spotIM.sso}
                 isUserSignedIn={spotIM.isUserSignedIn}
```
blech75: the `isUserSignedIn` prop was removed from the `SpotIM` component, so it should also be removed here, correct?

components/SpotIM/index.js
```diff
@@ -117,13 +117,22 @@ class SpotIM extends Component {
   }
 
   sso() {
-    window.SPOTIM.startSSO(this.props.sso)
+    const { communityId, auth0User } = this.props.user;
+    const site = process.env.SITE_KEY || 'k';
```
blech75: we should not rely on env vars in lib-ui. instead, the `site` should be passed in as a prop.

(also, `SITE_KEY` is not defined in the env, it's pulled from `constants/site` in oxy.)

components/SpotIM/index.js
```diff
@@ -153,11 +162,10 @@ class SpotIM extends Component {
     return (
       <ErrorBoundary>
         <Fragment>
-          {/* TEMP sign in link while we use Community based SSO */}
           {this.state.authorized === false && (
             <div className="SpotIM__login-link">
               <a
-                href={`/sign-in?return_to=${encodeURIComponent(
+                href={`/sign-in?redirect_uri=${encodeURIComponent(
```
blech75: it might be nice to DRY-up the login URL creation since it's used in a few places.

components/SpotIM/index.js
```diff
@@ -116,14 +116,28 @@ class SpotIM extends Component {
     return conversationScript;
   }
 
+  redirect() {
```
blech75: style nit: the private `redirect()` function you had it earlier was actually a-ok. this revised approach creates another instance method and (unnecessarily) binds it, which adds complexity to the class.

i envisioned the URL generation itself moving to a separate helper method; this class shouldn't be responsible for computing/storing it.

sorry for all of this; i just wanted to make my original intent clear! :-)

seawolff: @blech75 In order to define `redirectURL` as a reusable variable I need to set it in `componentDidMount` and because of that I had to move the redirect inside the component so it could reference that variable that is declared in the component life cycle.

seawolff: I'm actually surprised the other approached worked because I'm pretty sure it shouldn't have.

components/PostBundle/index.js
```diff
@@ -52,6 +52,9 @@ function PostBundle(props) {
     extraArticleClasses,
     itemPropMapperAdditions,
     pagePath,
+    user,
+    spotimIntegration,
+    site,
```
blech75: @seawolff getting a `'site' is missing in props validation eslint(react/prop-types)` warning.

blech75:
@seawolff getting a spec warning:
```
 PASS  components/SpotIM/index.spec.js
  ● Console

    console.error node_modules/prop-types/checkPropTypes.js:19
      Warning: Failed prop type: The prop `createSpotIMIntegration` is marked as required in `SpotIM`, but its value is `undefined`.
          in SpotIM
    console.error node_modules/prop-types/checkPropTypes.js:19
      Warning: Failed prop type: The prop `site` is marked as required in `SpotIM`, but its value is `undefined`.
          in SpotIM
```

seawolff:
@blech75 thanks for your help buddy!

PR #365, opened 2019-04-18
Henry/video kicker check

chenryhen:
Quick (I hope) review of this please. I initially tried to address the issue where teasers for standalone videos have no category, thus no kicker, here https://github.com/apartmenttherapy/oxygen-lib-ui/pull/364

However I realized belatedly that doing it at the teaser level would cause the `Video` kicker to show up, even if the bundle was set to `timestamp`. So now I'm trying to inject a check for a standalone teaser while setting bundle item properties.

The second bundle on K's `/collection/cheese` has a video teaser currently, so you can toggle the bundle's `kicker` setting in Workflow to check different conditions. When you expect to see a teaser, that video teaser should have a kicker of `Video`.

Still the same Asana card: https://app.asana.com/0/937985426519793/1119228747128132/f


chenryhen:
Adding/faking categories on the data from JWPlatform was apparently messing something up in streams? I'm fuzzy on the details, but the long term solution is to fix the incoming data, so the kicker value is available there, and we don't have to inject anything from the Oxy side.

PR #367, opened 2019-04-18
Replace moment with date-fns

seawolff:
+ This PR is an attempt to replace moment in oxy-lib-ui. Hopefully nothing was overlooked but I would love some extra eyes on this.

+ `Yarn test` should pass 
+ Confirm teasers on homepage have correct time
+ Confirm there is no weirdness with liftigniter
+ Test it live here: https://replacemoment-dot-oxygen-kitchn-dot-gen-prod.appspot.com/

+ Take a look at the following example (notice the time of the teaser "10 easy & classic easer recipes"):

__Moment__ (time: "An hour ago")
Screenshot 2019-04-18 15 23 07

__Date-FNS__ (time: "Just Now")
Screenshot 2019-04-18 15 23 18

Now the reason for this difference is between moment's `.diff` hourly comparison and date-fns' `differenceInHours`. In this particular case the time was approx. 37 minutes since the post had been published. In moment's case it will return that there is an hourly difference due to the fact it returns a float and rounds. But date-fns will simply compare the two dates and return a 0 or x hour differential. @andrewb is this correct? It seems odd to make posts seem older then they already are.

Screenshot 2019-04-18 15 22 49

You can check the breakdown of oxy-lib-ui by running `BUNDLE_ANALYZE=browser yarn build`

__origin/master__
Screenshot 2019-04-18 15 44 37

__wolff/replace-moment__
Screenshot 2019-04-18 15 43 45

https://app.asana.com/0/0/1114999684123473/f

* [x] add a reviewer
* [x] run appropriate tests
* [x] tag anyone who might want to see this


helpers/date.js
```diff
@@ -1,51 +1,51 @@
...
+  if (differenceInCalendarDays(new Date(), new Date(date)) < 1) {
     return subDay(date);
   }
   return day(date);
 }
 
 export function getIsoDate(seconds) {
   // ISO date with offset
-  return moment.unix(seconds).toISOString(true);
+  return new Date(seconds * 1000).toISOString();
```
andrewb: This won't give the same result. Moment's `toISOString(true)` prevents UTC conversion and preserves the offset.

We use this function for [LI metadata](https://github.com/apartmenttherapy/oxygen/blob/bbc1cf4318413b84eddf136fe74f2372d21481ec/components/Liftigniter/index.js#L37) and [video JSON-LD](https://github.com/apartmenttherapy/oxygen/blob/8c45c2ee3d336edfef4103d8863c0e832b171cf8/components/Video/index.js#L423). 

For example compare `published_time` in the `liftigniter-metadata` block for `/drop-dinners-267648`. On prod it is `2019-04-24T06:49:42.000-04:00` versus `2019-04-24T11:00:26.000Z`.

This doesn't necessarily have to change, and tbh using UTC is a lot simpler. What we need to do though is make sure it is ok to switch these to UTC. I'm not concerned about the JSON-LD for video – that will be fine. My only question is if it's ok to switch to UTC for LI? In theory it should be the same, but I don't know if the offset is just a hangover from ATCMS or if it is needed (and if there are any ramifications to changing it for indexed data). If finding an answer is difficult we could ask support. cc @blech75

andrewb:
@chenryhen what are your thoughts about the different language in the relative date strings?

helpers/date.js
```diff
@@ -1,51 +1,51 @@
...
 }
 
-export function getFormattedDate(seconds, format) {
+export function getFormattedDate(seconds, formatString) {
   // date in seconds formated via string
-  return moment.unix(seconds).format(format);
+  return format(seconds * 1000, formatString);
 }
 
 const DateHelper = {
```
andrewb: Not new to this PR, but we can fix the exports in this file?

```
const DateHelper = {
  getRelativeDate
};

export default DateHelper;
```

This feels odd since we're doing a named export of `getRelativeDate` anyway. We don't need a default export. If you don't mind fixing as part of this PR that'd be great.

chenryhen:
@andrewb @seawolff I actually should touch base with Maxwell on this as he was set on the shorter time descriptions. What is the actual trade-off here with date-fns? Better support and/or performance?

chenryhen:
@seawolff @andrewb managed to snag Maxwell. So we do want to keep it as short as possible, so is it an option, when things start getting very verbose (e.g. `about two months ago`) to just switch to the date? either something like 01/01/2019 or Jan 1, 2019?

andrewb:
@seawolff this is an existing problem, but something I'd love to solve with this PR is the issue with client and server dates not matching if the timezones are different. For example, I see the following warning if I change my timezone to something other than `America/New_York`:

```
Prop `data-gtm-teaser-promoted-at` did not match. Server: "2019-03-08 16:00 -0500" Client: "2019-03-08 13:00 -0800"
```

You can see this on the homepage.

@blech75 what's the purpose of `data-gtm-teaser-promoted-at`? Could we use a [UTC ISO string](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString) here instead?

seawolff:
@chenryhen how about once we get to a date over a month we switch to the formatted date you suggested of `01/01/2020`?

blech75:
@andrewb the value of `data-gtm-teaser-promoted-at` attr gets directly inserted into GA for reporting purposes. not sure how people use it or what would happen if we change it. could find out if needed.

the reason i'm using s custom format here is due to two reasons:

1) `.toISOString()` is slightly annoying to read with no spaces
2) it outputs in zulu/utc time and that is confusing for people to read/understand/translate since we base everything off `America/New_York`

`2019-03-08 16:00 -0500` (custom) vs. `2019-03-08T21:00:00.000Z` (via `.toISOString()`)

seawolff:
@blech75 Could you find out if we can change it to utc? I feel like it might help simplify some things as long as it doesn't interrupt how we're currently doing reporting. 

@andrewb odd that you get that error when switching timezones. Are you changing your system timezone?

chenryhen:
@seawolff could we make the cutoff sooner than a month or is that obnoxious? I think a short, readable format would be preferable (3-letter month, etc.)

seawolff:
@chenryhen I could do something like as soon as it's a day or week old we could show the date. Would that work? I need to be able to change it based on date range however, I don't think editing the format based on the string length is going to be reliable.

chenryhen:
@seawolff starting at a week old sounds good.

seawolff:
@chenryhen one more question. You only want to edit this format for the teasers, correct?

chenryhen:
@seawolff yup!

helpers/date.js
blech75: i emailed LI to see if changing timezone to UTC will cause issues for existing content. will report back.

i agree; using UTC for this purpose is simpler/saner.

blech75:
@andrewb i added a task for the timezone mismatch issue: https://app.asana.com/0/962855750175568/1120087925681831/f

@seawolff i'd prefer to keep `data-gtm-teaser-promoted-at` in the [custom format](https://github.com/apartmenttherapy/oxygen/blob/master/components/Teaser/index.js#L10) and NYC timezone if possible. we should be able to use [`format()`](https://date-fns.org/v1.30.1/docs/format) for this, though we'll likely still have the timezone mismatch issue above.

i'm going to branch off this PR and attempt to unpack the issue and devise a fix.

seawolff:
@blech75 Can we check that we need the timestamp in GA? I think the simplest fix might be just removing that entirely if it's not something that is needed by editorial.

seawolff:
btw I checked and I get this error now even on master. Most likely because I'm on central time

```
index.js:1 Warning: Prop `data-gtm-teaser-promoted-at` did not match. Server: "2019-04-30 09:45 -0400" Client: "2019-04-30 08:45 -0500"
```

blech75:
i made a comment in asana re: the timezone issue and confirmed it's _not_ related to this PR: https://app.asana.com/0/937985426519793/1120087925681831/f

looking into date formatting issue now...

helpers/date.js
blech75: forgot to mention. i heard back from LI and they confirmed the change to UTC is a-ok.

blech75:
ok, so i've confirmed: the only tangible difference that this PR introduces is the UTC change mentioned above: we now output datetimes for LI and video JSON-LD in UTC instead of `America/New_York` and this is a-ok per LI.

the value rendered to `data-gtm-teaser-promoted-at` does not change as a result of this PR: the format and value is identical so we're ok. no need to touch anything to do with that.

as i noted in the asana task referenced directly above, the issue with client-side datetime rendering not matching server display is a larger issue and has always existed in `master`; it needs to be solved independently of this PR. i have some ideas for that in the asana task.

helpers/date.js
blech75: @seawolff ping on this. just get rid of the lines cited above.

blech75:
@seawolff i _think_ the last step is to add `date-fns` as a dep to oxygen, right? (the lib-ui deps are not used during oxy's compilation.)

blech75:
@seawolff also you can remove this in oxy: https://github.com/apartmenttherapy/oxygen/blob/master/next.config.js#L78-L80

seawolff:
@blech75 corresponding oxy PR here https://github.com/apartmenttherapy/oxygen/pull/1381

PR #368, opened 2019-04-18
Fix SpotIM Sign Up to Post Button

seawolff:
This button is currently broken on prod. This PR adds an event listener so that this button can be clicked and the user can login/signip.

Screenshot 2019-04-18 16 39 38


components/SpotIM/index.js
```diff
@@ -45,12 +46,22 @@ class SpotIM extends Component {
     } else {
       this.onApiReady = this.logout;
       document.addEventListener('spot-im-api-ready', this.onApiReady, false);
+      document.addEventListener('spot-im-login-start', () => {
```
blech75: this event listener needs to be removed on unmount, too.

seawolff: @blech75 yep just realized that I didn't have that. Pushed.

components/SpotIM/index.js
```diff
@@ -35,6 +35,7 @@ class SpotIM extends Component {
       } else {
         this.onApiReady = this.sso;
         document.addEventListener('spot-im-api-ready', this.onApiReady, false);
+        document.addEventListener('spot-im-login-start', this.onApiReady);
```
blech75: missing the (annoying) last argument
```suggestion
        document.addEventListener('spot-im-login-start', this.onApiReady, false);
```

components/SpotIM/index.js
blech75: what i meant was: there's the possibility of two different functions attached to the `spot-im-login-start` event:

* the `this.onApiReady` method (in `if (isUserSignedIn) { ... }`.
* the inline method (in `if (window.SPOTIM && window.SPOTIM.logout) { ... }`). 

you need remove the correct function in each case.

seawolff: @blech75 Not sure I understand. The user can only be logged in or out and thus only one event listener will be bound at a time.

blech75:
i know you didn't initially work on this component, but it might be good to drop a reference to https://github.com/SpotIM/spotim-integration-docs/tree/master/api/js-events somewhere in this component. not sure where you found the magical `spot-im-login-start`, but that's where i found it.

components/SpotIM/index.js
seawolff: Oh I see what you mean. Nevermind. https://github.com/apartmenttherapy/oxygen-lib-ui/pull/368/commits/1488b39ab8472932ee9a1469f37a028a9bb71fbe

seawolff:
@blech75 There you go. Fixed.

components/SpotIM/index.js
```diff
@@ -8,6 +8,12 @@ const SPOT_LAUNCHER = 'https://launcher.spot.im/spot/';
 
 const NUM_PREVIEW_COMMENTS = 3; // Match what we used to have on ATCMS
 
+const redirect = () => {
+  window.location = `/sign-in?return_to=${encodeURIComponent(
```
blech75:
```suggestion
  // CLEANUP: use Router.push() when we switch to client-side routing.
  window.location = `/sign-in?return_to=${encodeURIComponent(
```
/cc @andrewb

seawolff:
@blech75 NP! Thanks for the quick review! 😄

PR #374, opened 2019-04-26
Add a fallback Category for a bundle's teasers and a filter for the categories

chenryhen:
SpreadBundle takes a fallback category for its teasers (although it gets a little complicated later on) as well as a whitelist of categories to check against. If the set category for the teaser isn't in the whitelist, just the `category` falls back to the site name (coming from Oxygen).

In the case of external links, we don't want the site name to prevent confusion, so the `Teaser` component checks for the presence of the attribute and adjusts the kicker accordingly. It's a little convoluted, but this is where it seemed to make sense with the chain of Oxy Spread > lib-ui Spread > Oxy Teaser > lib-ui Teaser

Concern:
The Teaser `icons.spec.js` doesn't check specifically for `<Play />` It looks like the child is just `mockConstructor`. How can I make that test better?


components/SpreadBundle/helper.js
```diff
@@ -126,10 +137,13 @@ export function createSetBundleItemProps(opts = {}) {
...
       }
 
+      const safeCategories = whitelist
+        ? filterTerms(data.categories, whitelist)
+        : data.categories;
       const category =
-        data.categories && data.categories.length > 0
-          ? data.categories[0].name
-          : '';
+        safeCategories && safeCategories.length > 0
```
blech75: it might be good to update the tests for the new logic of how a `category` gets computed.

components/Teaser/index.js
```diff
@@ -62,6 +62,11 @@ function Teaser(props) {
     hasLazyPicture
   };
 
+  const kicker =
+    baseKicker && teaserAttributes.includes('isExternalLink')
+      ? 'From the Web'
```
blech75: again, adding tests for this logic would be great, if you can. if not, no biggie

components/Teaser/icons.js
```diff
@@ -24,7 +23,8 @@ function TeaserIcons({ teaserAttributes }) {
 }
 
 TeaserIcons.propTypes = {
-  teaserAttributes: arrayOf(oneOf(Object.keys(ATTRIBUTE_ICONS))).isRequired
+  teaserAttributes: arrayOf(oneOf(['hasVideo', 'isExternalLink', 'isVideo']))
```
blech75: maybe i'm not understanding things correctly, but why would you allow `isExternalLink` here if the effective value of the used icon will be `undefined`? The previous approach handled only allowing the defined attrs/icons.

blech75:
i have an example of how to test for the exact (mocked) icon component. gimme a few to find it.

components/Teaser/icons.js
chenryhen: @blech75 I was getting an error because all of the `teaserAttributes` get passed into this `icons.js` file. If it's just `oneOf(['hasVideo', 'isVideo'])` React complains because `isExternalLink` isn't a valid prop value. 

I could just change this to `string` and then create a fallback for when the string isn't recognized as an icon match?

blech75: ah, ok. so you only want icons defined for certain attrs. you're gonna have to rework this a bit, then. something like this:
```js
const ATTRIBUTE_ICONS = {
  hasVideo: Play,
  isVideo: Play
};
```
and then
```js
{teaserAttributes
  .map(attribute => {
    const Icon = ATTRIBUTE_ICONS[attribute];
    return Icon ? (
      <div
        className={`Teaser__icon Teaser__icon--${attribute}`}
        key={attribute}
      >
        <Icon />
      </div>
    ) : null;
  })
  .filter(Boolean)}
```
strictly speaking, the first change -- ref'ing the component (`Play`)  instead of the element (`<Play />`) -- is not necessary, but i think it reads better and is slightly more efficient.

this will avoid outputting empty `<div>`s for missing icons.

you could also write this as a `.reduce()`, which might be slightly more efficient at the cost of a little readability for something so 'simple'.

blech75: i just realized i provided you with code but didn't answer your question about the props....

i might pull the list of allowed attrs out to a constant to make it clearer what you're doing.

```js
const ALLOWED_ATTRS = ['hasVideo', 'isVideo', 'isExternalLink'];
```
and then
```js
  teaserAttributes: arrayOf(oneOf(ALLOWED_ATTRS))
```

blech75:
@chenryhen this is how i might do the tests now:
```js
import { createSetupFunction } from '../../jest/helpers';
import TeaserIcons from './icons';
import { Play } from '../../constants/icons';

jest.mock('../../constants/icons', () => ({
  Play: jest.fn()
}));

const setup = createSetupFunction(TeaserIcons, {
  teaserAttributes: []
});

describe('<TeaserIcons />', () => {
  describe('sanity check', () => {
    it('should render', () => {
      const { wrapper } = setup();
      expect(wrapper.exists()).toBe(true);
    });
  });

  describe('uses attributes properly', () => {
    it('shows Play icon when `isVideo` flag is set', () => {
      const { wrapper } = setup({ teaserAttributes: ['isVideo'] });
      const iconEl = wrapper.find(Play);
      expect(iconEl.exists()).toBe(true);
    });

    it('should not output an icon for isExternalLink attribute', () => {
      const { wrapper } = setup({ teaserAttributes: ['isExternalLink'] });
      const divEl = wrapper.find('.Teaser__icon');
      expect(divEl.exists()).toBe(false);
    });

    it('shows only one icon when one valid flag is set', () => {
      const { wrapper } = setup({
        teaserAttributes: ['isVideo', 'isExternalLink']
      });
      const iconEl = wrapper.find('.Teaser__icon');
      expect(iconEl).toHaveLength(1);
    });
  });
});
```

components/SpreadBundle/helper.js
```diff
@@ -71,10 +73,18 @@ export function extrapolateExceptions(exceptionPatterns, totalItems) {
  * @param {Object} opts.bundleData
  * @param {String} opts.bundleData.kicker
  * @param {Object[]} opts.bundleData.categories
+ * @param {string} opts.fallbackCategory
```
blech75: i'm sorry, do you mind doing me a favor doing a quick var rename?
* `fallbackCategory` → `categoryFallback`
* `whitelist` → `categoryWhitelist`

i just want to make it clearer that we're dealing with category data here, and that they're related.

components/Teaser/index.js
```diff
@@ -62,6 +64,11 @@ function Teaser(props) {
     hasLazyPicture
   };
 
+  const kicker =
+    baseKicker && teaserAttributes.includes('isExternalLink')
```
blech75: you can use `IS_EXTERNAL_LINK_ATTR` here.

PR #375, opened 2019-04-26
hide key post elements on WP pages

blech75:
courtesy PR. i don't like this, but it will satisfy the requirements right now.


blech75:
gah, i should probably add specs for this.

PR #388, opened 2019-05-17
Henry/css placeholder

chenryhen:
This turned into a much more lib-ui heavy approach than I originally anticipated. I was going to add placeholder styling into each variant's css file, but this was certainly faster for me. 

Replaces the placeholder image in each teaser with a component that renders and empty div and a `<style>` block that uses the teaser's picture's dimensions to style the block into the correct size and shape. 

I recommend grabbing oxygen https://github.com/apartmenttherapy/oxygen/pull/1423 if you actually want to see the placeholder, otherwise it's transparent.

Take a look at any collection page and you should see that each teaser has a correctly-sized area blocked off for the image before the image loads. It will help to throttle the connection, or just scroll down the page really fast. 

If you change your browser size the placeholders should resize appropriately.

So soooooometimes, more than one teaser placeholder gets the same instance `_id` counter somehow. I think it's related to bundle loading and when the placeholder component is being used. I believe that a more random/unique id for each instance could fix this.

This adds a ton of `<style>` blocks to the markup. I hope the trade-off with not needing the image request will make up for this. Each SpreadBundle variant does have its own config file for images (e.g. `variantSoundbar.js`), so there's probably a way to set styles for the placeholders at the bundle level dynamically. If we don't get some speed wins with _this_ change, maybe that could be the next step.

#### GIF
parking-spot-saver-360

* [ ] add a reviewer
* [ ] run appropriate tests


chenryhen:
Actually, I totally forgot that all of that placeholder `<picture>` markup is gone, so the page will probably end up smaller than before. 🎇

chenryhen:
FYI, I still don't know how to reproduce it, but the duplicate placeholder `_id` is happening for me still. It'll probably be best to generate a uuid or some other unique id a different way.

components/TeaserImage/placeholder.js
```diff
@@ -0,0 +1,44 @@
...
+    this._id = _id;
+    _id += 1;
+  }
+
+  render() {
+    const { sourceSets } = this.props;
+    const cssRules = createCSSRules(sourceSets, this._id).reverse();
+    return (
+      <Fragment>
+        <style dangerouslySetInnerHTML={{ __html: cssRules.join(' ') }} />
```
andrewb: @chenryhen oh, I think there will be an easier way when we migrate lib-ui into the core. I think you'll be able to use styled-jsx for this instead which will scope it to the component, which also means you won't need to worry about the ids either.

chenryhen: @andrewb should I put this on hold until then? Or do we want to see if there's any performance win first?

